PR #11, opened 2025-12-28
local-arp-tcp branch fixes

ddirect:
Some fixes needed to make the local-arp-tcp branch functional with examples/xnet. A couple of them are from my original PR.
I guess the main thing to do is deciding how to handle the issue with the ARP connection ID (see https://github.com/soypat/lneto/pull/8#pullrequestreview-3614738208).

### DIFF
--- a/arp/handler.go
+++ b/arp/handler.go
@@ -255,7 +255,7 @@ func (h *Handler) Demux(ethFrame []byte, frameOffset int) error {
 }
 
 func trySetEthernetDst(ethFrame []byte, dst []byte) {
-	if len(ethFrame) > 14 {
+	if len(ethFrame) >= 14 {
 		copy(ethFrame[:6], dst)
 	}
 }
--- a/x/xnet/stack-async.go
+++ b/x/xnet/stack-async.go
@@ -28,7 +28,7 @@ type StackAsync struct {
 	ip       internet.StackIP
 	arp      arp.Handler
 	udps     internet.StackPorts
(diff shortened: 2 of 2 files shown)


arp/handler.go
```diff
@@ -43,7 +43,7 @@ func (h *Handler) Reset(cfg HandlerConfig) error {
 		return errors.New("invalid Handler query or pending config")
 	}
 	*h = Handler{
-		connID:          h.connID + 1,
+		connID:          h.connID,
```
ddirect: This is temporary, just to make things work (see https://github.com/soypat/lneto/pull/8#pullrequestreview-3614738208)

soypat: So a reset should trigger a disconnect, to fix the behaviour in StackASync where the ARP node is lost i committed: https://github.com/soypat/lneto/pull/10/changes/a2886423a2fd4668811328588aad05a402c66321 

However, as you say, disconnecting in a DHCP assimilation is a bug, so a fix for that could look like: https://github.com/soypat/lneto/pull/10/changes/890fd33b22b98c80b3da9c4140b689dae8c3a60c

Let me know what you think

soypat:
I've added 2 commits on top of local-arp-tcp that should fix the connectionID issue- please test if that works for you. As for the rest of your changes, look great. As soon as you test I'll see we merge all of them

soypat:
I'm gonna merge and touch things up a bit to see if we can speed up merge of the whole :)

PR #12, opened 2025-12-30
Fix Local ARP

soypat:


soypat:
@ddirect I've added a test that verifies the ARP for a IP address on the subnet is resolved

ddirect:
@soypat I tested this PR as it is now with examples/xnet and it seems to work fine

PR #18, opened 2026-01-12
arp: fixes to implementation

egonelbre:
* typo in error text
* compactQueries was keeping invalid queries
* compactQueries accidentally ended up creating shared slices


soypat:
Thank you for your contribution Egon!

PR #19, opened 2026-01-13
dhcpv4: fixes

egonelbre:
* condition flipped in Encapsulate
* check for correct bounds in ForEachOption
* getMessageType didn't check for opt
* reset backing buffers


soypat:
Yet again, very neat finds and improvements.

PR #20, opened 2026-01-18
Repeating select call on Bridge when interrupted

ddirect:
I had some trouble with the select call used to wait data on the raw socket ("Bridge") which lneto examples use under Linux. It seems that the operation is interrupted from time to time on my bare metal Ubuntu 24.04, and in those cases it needs to be repeated.
Note that the "Tap" select call may need to be patched too.


soypat:
I've added retry limit, is this OK by you?

soypat:
Thank you again for the imrpovement ddirect!

PR #30, opened 2026-01-27
TCP Conn Write backoff used to wait for the TX buffer to have enough room

ddirect:
@soypat Just a proposal, I thought it was faster to show you directly with a PR.
We discussed about using `Conn.AvailableOutput()` to determine if it's possible to send data on a TCP Conn without getting the `buffer full` error. After working on backoff, I realized that Conn.Write already has a mechanism to do this (the backoff). But the ring buffer needs to return (0, nil) (no error) in order to have the backoff mechanism being activated.

This change works fine in my tests and allows me to drop the `Conn.AvailableOutput()` check before writing. 

Alternatively, we could continue to return `errRingBufferFull` from the ring buffer, and detect that condition inside Conn.Write, to activate the backoff.

### DIFF
--- a/internal/ring.go
+++ b/internal/ring.go
@@ -9,8 +9,7 @@ import (
 )
 
 var (
-	errRingBufferFull = errors.New("lneto/ring: buffer full")
-	errRingNoData     = errors.New("lneto/ring: empty write")
+	errRingNoData = errors.New("lneto/ring: empty write")
 )
 
 // Ring implements basic Ring buffer functionality.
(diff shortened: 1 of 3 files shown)


soypat:
should be hit when n==0... which is never hit now you mention it. Hmmm not sure what the ideal pattern is going forward. Pretty sure this is a high-level-enough-call to warrant sleeping if the buffer is full. Maybe we need to add an early check for fullness and then sleep if buffer is full- or if the write returns ring buffer full error.

PR #43, opened 2026-02-27
Error rewrites

soypat:
Needs more logging to cover for context lost due to simpler errors.


soypat:
Heck, we're merging now since lneto is advancing quite fast. we can add logging and fix awkward use of errors in future when we run into the pain of using a limited error subset. We can then choose to expand error set to include more errors and get better error messages or add logs to complement errors. priority right now is letting me test all these changes integrated on the LAN8720

Issue #47, opened 2026-02-28
On the latest version, after 3-5 minutes, i receive the error seq not in snd/rcv.wnd


RaresAil:
With version `v0.0.0-20260227153846-eab43c4653d7` after 3-5 minutes i receive this error `reject in/out seg: seq not in snd/rcv.wnd` and when publishing messages i receive `i/o timeout`. I had to rollback to `v0.0.0-20260225155855-be949d960c6b` in order to fix the issue

soypat:
Thank you for the report, will investigate

PR #49, opened 2026-03-01
tcp: bugfix for seq not in snd/rcv.wnd error

soypat:
Fixes https://github.com/soypat/lneto/issues/47

### DIFF
--- a/tcp/conn.go
+++ b/tcp/conn.go
@@ -111,18 +111,21 @@ func (conn *Conn) BufferedUnsent() int {
 	return conn.h.BufferedUnsent()
 }
 
-func (conn *Conn) AvailableInput() int {
+// FreeInput returns the number of free bytes in the socket's receive(input) buffer.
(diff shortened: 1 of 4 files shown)


soypat:
Go to your project root directory and run

`GOPROXY=direct go get -u github.com/soypat/lneto@latest`

Tells go to upgrade lneto dependency to latest by getring it from the link directly and not using the go module cache which may have an older version.

soypat:
Found SO many divergences from spec- but I'm still not quite sure I've fixed your particular issue. Will look at your logs next

RaresAil:
Thank you, i will check your example and come back with info to see how that one performs, the current example that i used was in the [cyw43439](https://github.com/soypat/cyw43439/blob/main/examples/mqtt/mqtt.go) maybe the one from there should be updated as well

RaresAil:
i will create another issue for this

Issue #36, opened 2026-02-09
Network timing on TCP


soypat:
Need to eventually add timing of packets to enable retransmission of TCP packets after the other side fails to get them. 

- Likely this will mean adding a `Time func() time.Time` field to TCP configuration, if nil will use standard library time- or breaking change and make sure user sets this?
- Thinking we should provide the xnet.StackAsync "now" time field which is set whenever Encapsulate/Demux is called, the TCP conns all get this time to avoid hammering the hardware with time.Now. 
- Modify tests to use a test time source to avoid all flakyness. Idea: Should be incremented by 1ns every encapsulate/demux

## Alternatives
- Store time in an uint32 representing microseconds (TCP clock as per RFC), atomic access to read the time. Can pass in a pointer to uint32 to TCPConn, is updated every demux/encapsulate call
- Time as a parameter to Demux and Encapsulate. Though this seems like a waste of stack memory, and a large one at that.

ddirect:
Hi Pato, I have a question which kind of relates with this: I see that _ringTx_ keeps the details of all sent TCP packets (position, sequence and size) in a _sentlist_ instance. I would think that the indexes of the sent/unsent/acked(=free) regions of the ring buffer, together with the sequence counter and (at least one) timestamp, would be enough to keep track of what needs to be sent or retransmitted. Why did you choose to keep the history of all sent packets?

soypat:
> Currently, Handler.Send attempts to send as many packets as the TX buffer and packet queue allows. Instead, before doing this we need to check how long the first packet in the packet queue has been in flight, and, if that's above some threshold, that packet should be resent.

So my understanding is we have two options in this situation- We either 1. Prioritize filling the window or 2. Prioritize sending stale unacked packets. I do tend to agree sending prioritizing stale packets sounds better.

> We can add transmission timestamps to each packet in the queue to keep track of the in-flight time.

Yes, sounds good. I wonder if we should store them as time.Time (128 bits) or as 32 bit microsecond timestamps, and have the tcp.Clock type be `func() uint32`. I think the spec says something about requiring microsecond level accuracy and 32bit being sufficient, will double check this. 

> As for the threshold, we should attempt to make an estimate of the current network round-trip, and multiply it by some factor to keep some margin.

We can keep track of last 2..4 acksand calculate a average network round trip time? We should totally investigate how linux, Zephyr, smoltcp do this and choose what best suits us.

> "retransmit mode"

[...]

PR #52, opened 2026-03-04
Implement basic retransmit queue reset method

soypat:
Solves https://github.com/soypat/lneto/issues/36 with a very basic strategy. Upon fulfilling condition for retransmit we simply reset the sent queue so as to act as if the already sent packets are unsent. This will retransmit these packets as if they were being transmitted for the first time


RaresAil:
Alright, thank you, the only issue i found with cores is only the flash, i started 6 hours ago the device with cores and everything works perfectly

PR #74, opened 2026-04-11
add corpus

soypat:
Mostly just curious as to how a corpus changes code coverage, not sure about adding it


soypat:
Nah, does not seem worth a few percent

PR #84, opened 2026-04-15
Replace go.yml with comprehensive ci.yaml

MDr164:
- Rename go.yml to ci.yaml
- Pin all actions to full commit SHAs (checkout v6.0.2, setup-go v6.4.0, codecov v6.0.0, setup-tinygo v2.0.1, github-action-benchmark v1.22.0)
- Add dedicated benchmark job with regression detection via benchmark-action/github-action-benchmark (120% alert threshold)
- Set workflow-level permissions: read-all with per-job write grants


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version: "1.25"
+
+      - name: Run benchmarks
+        run: go test -bench=. -benchmem -count=5 -shuffle=on -run='^$' -timeout=15m ./... | tee bench-results.txt
+
+      - name: Report benchmark regressions
+        uses: benchmark-action/github-action-benchmark@a60cea5bc7b49e15c1f58f411161f99e0df48372 # v1.22.0
```
soypat: What is `benchmark-action/github-action-benchmark`? Is there a marketplace page for it?

MDr164: This one: https://github.com/benchmark-action/github-action-benchmark

MDr164: I could also drop the benchmark analysis action. I made it non-critical at least so it will pass even on a regression in case you prefer having a :heavy_check_mark: instead of a red cross there. Other than that no 3rd party actions have been added.

MDr164:
Btw. I omitted running any fuzz testing in CI for now. I worked with oss-fuzz in the past but setting it up is kind of clunky. Another approach would be to just run nightly for 30min and propagate the seed corpus between runs or so.

.github/workflows/ci.yaml
soypat: Would it be OK to use v1.21 instead of the fresher v1.22 which is 2 weeks old. I've learned to profoundly distrust the npm ecosystem.

MDr164: Sure thing, will update the commit. That's also why I pinned every action to a commit instead of a version like it was done in the old yaml file. The versions could simply point to a different commit and silently replace the executed action code.

MDr164:
I'll drop the memprofile and cpuprofile flags, I wasn't aware that these can't be used with the shorthand `./...` call of go test. And I don't feel like writing a for loop in the yaml. I suppose one can add the flags when trying to analyze the code locally and we already got the benchmarks anyways.

soypat:
Why did you not use `-newexpr` in go fix? Or is that for a future PR? 

Seems like benchmarks are failing. Can we add benchmark regressions in a future PR? This one is quite loaded as is. I also would not mind if we could find a way to avoid NPM :)

MDr164:
> Why did you not use `-newexpr` in go fix? Or is that for a future PR?
> 
> Seems like benchmarks are failing. Can we add benchmark regressions in a future PR? This one is quite loaded as is. I also would not mind if we could find a way to avoid NPM :)

I had a look and because my local Go is 1.26 the syntax for `go fix` is slightly different. I figured it makes sense to not only call it with newexpr but run all analyzers (which is the default if none is specified). Also the -fix -diff combo is something that had been added in 1.26 it seems. I'd prefer to bump the Go version used in CI to 1.26 as it gives better tooling and should not affect the code at all as it still honors the `go.mod` file.

And yes, I'll drop the thirdparty action. Benchmarks are still run but no reporting happens then. One would just need to click on the action and look into the logs.

MDr164:
Ok I'm kind of clueless why that step fails in CI but works locally. I'll investigate and update the lint task.

EDIT: Got it I think, `go tool fix` vs `go fix`. Both work locally but the docs mention that former should not be used. Looks like we found out why.

MDr164:
Nice, now everything works! The lint failures are being addressed in #85

PR #85, opened 2026-04-15
Bump minimum Go version to 1.24

MDr164:
I updated to 1.25 (currently 1.26 is stable but 1.25 is more widely available I suppose). As newer Go version simplified some syntax I simply ran `go fix -newexpr ./...` which will call the go modernize tool. Hence all the code changes have been done directly by the go tool and not manually. I'm not too sure which version tinygo is still compatible with as I'm not so familiar with that toolchain.

### DIFF
--- a/internal/ltesto/ltesto.go
+++ b/internal/ltesto/ltesto.go
@@ -489,11 +489,8 @@ func mutateTCPOptions(opts []byte, seed int64, variant int64) int64 {
 		}
 	case 4: // SACK with garbage block data.
 		if len(opts) >= 10 {
-			opts[0] = byte(tcp.OptSACK) // kind=5
-			sackLen := len(opts)
-			if sackLen > 34 {
-				sackLen = 34 // max 4 SACK blocks
-			}
+			opts[0] = byte(tcp.OptSACK)   // kind=5
+			sackLen := min(len(opts), 34) // max 4 SACK blocks
 			opts[1] = byte(sackLen)
 			for i := 2; i < sackLen; i++ {
 				opts[i] = byte(seed)
--- a/dhcpv4/client.go
+++ b/dhcpv4/client.go
@@ -153,10 +153,7 @@ func (c *Client) Encapsulate(carrierData []byte, offsetToIP, offsetToFrame int) 
 		numOpts += n
 		n, _ = EncodeOption(opts[numOpts:], OptParameterRequestList, defaultParamReqList...)
 		numOpts += n
-		maxlen := len(dst)
-		if maxlen > math.MaxUint16 {
-			maxlen = math.MaxUint16
-		}
+		maxlen := min(len(dst), math.MaxUint16)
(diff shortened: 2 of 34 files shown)


soypat:
Nice! If we are going to modernize the codebase I'd like to have a check in CI that fails on PRs that have changes applied on -newexpr, or else we're setting up to running newexpr every once in a while

soypat:
Also I'd want to support as far back as possible. the min function was added 1.21 and `for range 10` was added 1.22. So unless there's a good reason to bump up from 1.23 I'd stick with same go.mod

MDr164:
The reason why I would have liked to at least go to 1.24 is `testing.B.Loop()`: https://go.dev/blog/testing-b-loop

That's currently the only reason I had for moving past 1.23

MDr164:
> Nice! If we are going to modernize the codebase I'd like to have a check in CI that fails on PRs that have changes applied on -newexpr, or else we're setting up to running newexpr every once in a while

There's two ways to go about that imo, either run `go fix -newexpr` and check for output, similar to how I did it in the CI PR with gofmt or we could add a very targeted golangci-lint config (e.g. only run a very minimal set of linters like go modernize, gofmt, go vet, etc). That would require to set up golangci-lint as another CI tool. I'm fine with doing that but let me know if you got a preference there.

soypat:
The simpler the better- if go fix -newexpr works then lets go for it!

Edit: Alright, I like adding support for b.Loop(). We can bump to 1.24

MDr164:
I'll change to 1.24 then :)

Should I add the CI additions here in https://github.com/soypat/lneto/pull/84 while it's still open?

EDIT: Updated the CI PR with a `go fix` invocation.

internal/ltesto/ltesto.go
```diff
@@ -490,10 +490,9 @@ func mutateTCPOptions(opts []byte, seed int64, variant int64) int64 {
 	case 4: // SACK with garbage block data.
 		if len(opts) >= 10 {
 			opts[0] = byte(tcp.OptSACK) // kind=5
-			sackLen := len(opts)
-			if sackLen > 34 {
-				sackLen = 34 // max 4 SACK blocks
-			}
+			sackLen := min(len(opts),
+				// max 4 SACK blocks
```
soypat: That looks funny, mind setting the comment and code in one line?

MDr164: Oh true. I fixed that now. Looks like `go fix` dealt with the comment weirdly when using the `min()` builtin

soypat:
@MDr164 Thank you for your contributions!

MDr164:
> @MDr164 Thank you for your contributions!

Thanks for the reviews. With the two 'housekeeping' PRs in I'll go ahead with some actual code. Currently trying to figure out the best way to validate that the protocols I'll add actually adhere to the spec without too much manual work in wireshark.

PR #88, opened 2026-04-17
Add NTS support

### LINKED ISSUE #86: NTS support
RFC: https://www.rfc-editor.org/rfc/rfc8915

lneto currently supports basic NTP but I already spotted a couple shortcomings in the implementation and other NTP clients like chrony for example also ships with NTS (NTP over TLS) which is also supported by several NTP servers already. It extends NTP and requires two subprotocols (NTS-KE and NTS which adds NTP extensions fields). While NTS-KE simply mandates TLS 1.3 which could be pulled in from stdlib crypto the actual NTP extension requires the use of at least one specific AEAD algorithm which is AEAD_AES_SIV_CMAC_256 according to the RFC. This one could either be the users obligation to provide using the `cipher.AEAD` interface or implemented in here as a minimal pure-go variant. Either way I'd propose adding NTS under the `x` package for now.

I would assign myself to the issue but got not enough permissions to do so.

MDr164:
This PR adds full NTS support and has been validated against cloudflares NTS servers. There are still a couple items in the code that I'm not fully contempt with especially pulling a crypto package into this repo. I'll keep this PR a draft until I deem the code ready. This resolves #86


ntp/client.go
```diff
@@ -61,28 +68,40 @@ func (c *Client) Encapsulate(carrierData []byte, offsetToIP, offsetToFrame int)
 		return 0, err
 	}
 
+	var exchange int
```
soypat, line 71: Remove exchange and use c.state to determine actual state and what message was received.

ntp/client.go
```diff
@@ -101,17 +120,49 @@ func (c *Client) Demux(carrierData []byte, frameOffset int) error {
...
 		}
-
 		txelapsed := c.now().Sub(c.start)
 		c.t[1] = frm.ReceiveTime()
 		c.t[2] = xmt
 		c.t[3] = c.t[0].Add(txelapsed)
 		c.serverStratum = frm.Stratum()
-		c.state = stateDone // TODO: add second exchange part.
+		c.offset1 = (c.t[1].Sub(c.t[0]) + c.t[2].Sub(c.t[3])) / 2
+		c.rtt1 = c.t[3].Sub(c.t[0]) - c.t[2].Sub(c.t[1])
```
soypat, line 135: Add comments referencing sections, names, methodology of the NTP RFC which are not explicit in the code. It's been a while since I've read up on it and I did not leave much documentation myself. I recall there being t[0..3] and thought that was all you needed to calculate the values of interest.

ntp/client.go
```diff
@@ -101,17 +120,49 @@ func (c *Client) Demux(carrierData []byte, frameOffset int) error {
...
+				slog.Duration("offset", c.offset1), slog.Duration("rtt", c.rtt1),
+				slog.String("stratum", c.serverStratum.String()))
+		}
 	case stateAwait2:
+		xmt := frm.TransmitTime()
+		orig := frm.OriginTime()
+		if xmt == orig || orig != c.t[0] {
+			if c._log != nil {
+				c._log.Debug("ntp.Client:demux:drop", slog.String("reason", "origin mismatch"),
+					slog.Uint64("orig.sec", uint64(orig.Seconds())), slog.Uint64("T1.sec", uint64(c.t[0].Seconds())))
```
soypat, line 148: Would be better to add a Uint64LittleEndian or Uint64le (or bigendian, whatever makes time increase it monotonically) method on Timestamp that returns full value of the timestamp for printing and encoding

ntp/client.go
```diff
@@ -101,17 +120,49 @@ func (c *Client) Demux(carrierData []byte, frameOffset int) error {
...
+			}
+			return lneto.ErrPacketDrop
+		}
+		txelapsed := c.now().Sub(c.start)
+		c.t[1] = frm.ReceiveTime()
+		c.t[2] = xmt
+		c.t[3] = c.t[0].Add(txelapsed)
+		c.serverStratum = frm.Stratum()
```
soypat, line 156: stratum was set in first exchange.

ntp/client.go
```diff
@@ -101,17 +120,49 @@ func (c *Client) Demux(carrierData []byte, frameOffset int) error {
...
+		offset2 := (c.t[1].Sub(c.t[0]) + c.t[2].Sub(c.t[3])) / 2
+		rtt2 := c.t[3].Sub(c.t[0]) - c.t[2].Sub(c.t[1])
```
soypat, line 158: Put these inside of the if statement

ntp/extensions.go
```diff
@@ -0,0 +1,115 @@
...
+// NextExtField parses the first NTP extension field from buf and returns it along
+// with the remaining unprocessed bytes. An empty buf returns io.EOF-equivalent nil
+// error with an empty ExtField. Use this in a loop:
+//
+//	field, rest, err := ntp.NextExtField(payload)
+//	for err == nil && len(field.RawData()) > 0 {
+//	    // process field
+//	    field, rest, err = ntp.NextExtField(rest)
+//	}
+func NextExtField(buf []byte) (field ExtField, rest []byte, err error) {
```
soypat, line 76: Instead of returning rest return the amount of bytes consumed here.

ntp/handler.go
```diff
@@ -0,0 +1,133 @@
...
+	RefID      [4]byte
+	MaxPending int
+}
+
+// Handler is a basic NTP server implementing [lneto.StackNode].
+// It receives client requests via [Handler.Demux] and builds server
+// responses via [Handler.Encapsulate].
+//
+// Handler is not safe for concurrent use.
+type Handler struct {
```
soypat, line 23: Why not call it Server?

ntp/ntp.go
```diff
@@ -131,6 +131,37 @@ func (frm Frame) SetTransmitTime(rt Timestamp) {
 	rt.Put(frm.buf[40:48])
 }
 
+// RawData returns the underlying byte slice for the entire NTP packet.
+func (frm Frame) RawData() []byte { return frm.buf }
+
+// Payload returns the extension fields area of the NTP packet (all bytes
+// following the fixed 48-byte NTP header). Returns nil for packets with no
+// extension fields.
+func (frm Frame) Payload() []byte {
```
soypat, line 140: Is Payload a good name? Wikipedia says what follows the fixed header are optional items. Do options sound better? Honest question btw, does the RFC call them something?

ntp/ntp.go
```diff
@@ -131,6 +131,37 @@ func (frm Frame) SetTransmitTime(rt Timestamp) {
...
+	if len(frm.buf) <= SizeHeader {
+		return nil
+	}
```
soypat, line 143: Remove `if`. It's OK to panic. Frames need to be constructed with NewFrame, if they are not they are invalid frames. This is how all protocols are designed in lneto. Never had a panic

x/nts/nts_test.go
```diff
@@ -0,0 +1,733 @@
...
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/binary"
+	"math/big"
+	"net"
+	"testing"
+	"time"
+
+	"golang.org/x/crypto/chacha20poly1305"
```
soypat, line 17: It seems that chacha is only used in tests- why is it necessary if it is not part of the main implementation?

soypat:
Regarding having LLMs be authors I think we stand with the Go team's recent approach/position. Have a note in the commit message noting the LLM that assisted but do not include them as authors. You as a developer should review all of the code and be in full knowledge of what is being put forth. In the words of Russ "As far as I can tell having an LLM in the authors list is just free publicity".

x/nts/nts_test.go
soypat: Heads up! Most recent commits is how one can include imports without impacting the project go.mod FYI, maybe we can have a test pipeline that uses imports- but if we can avoid imports in Lneto and the CI then better by me. https://github.com/soypat/lneto/tree/main/examples/_import_examples/gvisor-mwe

ntp/handler.go
MDr164: No particular reason, will rename to make the intent clearer. I'm just more used to calling things `Handler`

ntp/ntp.go
MDr164: Maybe not the best name, no.  RFC 7822 §2 uses "extension fields" terminology which I already had in the comment, I assume I'll just call it ExtensionFields then and don't try to summarize its data under the too generic "Payload" term :+1:

x/nts/nts_test.go
MDr164: I wanted to use a more common AEAD algorithm for testing (that one is not mandatory to use but at least in x/crypto) in order to make sure the AEAD interface is correct but in hindsight this test isn't all that useful so instead of creating another go.mod file I'll just drop this test. I'm still not really contempt with keeping the SIV algorithm in here either actually, any opinions on keeping or dropping it? It is required by the RFC but it is a crypto algorithm which this project really isn't about.

MDr164:
The PR got a tad bit large, I'm wondering if I should split this into smaller chunks, e.g. one to extend the NTP protocol with extensions fields, one that adds the client/server examples, one that adds the crypto code (which I'm still not really a fan of having it here) and one that ultimately adds NTS. It's already split into commits inside this PR but maybe it's worth splitting into several PRs.

examples/ntp-client/main.go
```diff
@@ -0,0 +1,83 @@
...
+	flag.Parse()
+
+	conn, err := net.DialTimeout("udp", *addr, 5*time.Second)
+	if err != nil {
+		return fmt.Errorf("dial: %w", err)
+	}
+	defer conn.Close()
+
+	var client ntp.Client
+	client.Reset(-18, time.Now)
```
soypat, line 47: -18? Could we use the ntp package sysprec provided function?

examples/ntp-client/main.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	for !client.IsDone() {
+		reqBuf := make([]byte, ntp.SizeHeader)
+		n, err := client.Encapsulate(reqBuf, 0, 0)
+		if err != nil {
+			return fmt.Errorf("encapsulate: %w", err)
+		}
+		if n == 0 {
+			continue
```
soypat, line 59: This n==0 -> continue is odd. if this branch is hit it will loop forever if the client is not done. add a fixed number of retries to the for loop or something.

examples/ntp-server/main.go
```diff
@@ -0,0 +1,83 @@
...
+		return fmt.Errorf("listen: %w", err)
+	}
+	defer pc.Close()
+	fmt.Printf("NTP server listening on %s\n", pc.LocalAddr())
+
+	var handler ntp.Server
+	err = handler.Reset(ntp.ServerConfig{
+		Now:        time.Now,
+		Stratum:    ntp.StratumPrimary,
+		Precision:  -20,
```
soypat, line 50: ntp provided system prec function not useful?

examples/ntp-server/main.go
```diff
@@ -0,0 +1,83 @@
...
+		RefID:      [4]byte{'G', 'O', 'L', 'N'},
+		MaxPending: 16,
+	})
+	if err != nil {
+		return fmt.Errorf("server reset: %w", err)
+	}
+
+	buf := make([]byte, 1500)
+	for {
```
soypat, line 59: maybe should include a backoff since this loop would consume 100% of CPU if a no-wait branch is hit repeatedly? i.e: backoff reset after sending/receiving, backoff incremented after hitting no rxdata/no txdata sent. Look at uses of lneto.BackoffStrategy in this repo.

examples/nts-client/main.go
```diff
@@ -0,0 +1,184 @@
...
+	defer conn.Close()
+
+	for !client.IsDone() {
+		carrier := make([]byte, 1500)
+		n, err := client.Encapsulate(carrier, 0, 0)
+		if err != nil {
+			return fmt.Errorf("encapsulate: %w", err)
+		}
+		if n == 0 {
+			continue
```
soypat, line 155: Same two points I mentioned in server and client examples. also something I noted in all examples is you could use a single buffer for sending/receiving allocated before the loop, but instead allocate a new one(or two) on every loop. this second point is not critical but would be a nice to have since in lneto we are in the business of avoiding heap allocations. Especially if we are going to use LLMs we should maintain a homogenous standard to not confuse them

ntp/client.go
```diff
@@ -62,27 +69,32 @@ func (c *Client) Encapsulate(carrierData []byte, offsetToIP, offsetToFrame int)
...
-	for i := range payload[:SizeHeader] {
-		payload[i] = 0
-	}
-
 	frm.ClearHeader()
 	frm.SetStratum(StratumUnsync)
 	frm.SetPoll(6)
 	frm.SetPrecision(c.sysprec)
-	frm.SetOriginTime(c.t[0])
+	frm.SetTransmitTime(c.t[0])
```
soypat, line 92: According to the documentation I've written on both methods this should be SetOriginTime (?)

soypat: even more so if below we do orig != c.t[0]

ntp/client.go
```diff
@@ -97,21 +109,59 @@ func (c *Client) Demux(carrierData []byte, frameOffset int) error {
...
 
-		txelapsed := c.now().Sub(c.start)
-		c.t[1] = frm.ReceiveTime()
-		c.t[2] = xmt
-		c.t[3] = c.t[0].Add(txelapsed)
+	// RFC 5905 §8 on-wire protocol: compute T1..T4 timestamps.
+	//  T1 = c.t[0] (client transmit, captured in Encapsulate)
+	//  T2 = ReceiveTime (server receive)
+	//  T3 = TransmitTime (server transmit)
+	//  T4 = T1 + elapsed local time since Encapsulate
```
soypat, line 136: These 4 lines are already documented in the struct (except for T4 expression?); I'd make the comment more concise since I don't expect readers to be in the dark about what the data structures at play are. That much we can expect readers to be familiar with

ntp/client.go
```diff
@@ -97,21 +109,59 @@ func (c *Client) Demux(carrierData []byte, frameOffset int) error {
...
+				slog.String("stratum", c.serverStratum.String()))
+		}
+	} else {
 		c.state = stateDone
+		if c._log != nil {
+			c._log.Debug("ntp.Client:demux:exchange2",
+				slog.Duration("offset", offset), slog.Duration("rtt", rtt),
+				slog.Duration("avg_offset", (c.offset1+offset)/2),
+				slog.Duration("avg_rtt", (c.rtt1+rtt)/2))
+		}
```
soypat, line 164: I think the Debug/Info/Error etc. family of log methods allocate in some cases (or all cases). Either use LogAttrs or follow the pattern used in the other protocol loggers. See tcp/debug.go for one such example.

I wonder if adding a internal.Logger type might be nice to have

```go
package internal
type Logger struct {
   log *slog.Logger
}
func (l *Logger) Info(msg string, attrs ...slog.Attr) {
    LogAttrs(l, slog.LevelInfo, msg, attrs...) // This is already a pkglevel func.
}
// ... etc
```

ntp/client_test.go
```diff
@@ -29,11 +53,7 @@ func TestClient_FullExchange(t *testing.T) {
 		t.Fatalf("expected %d bytes, got %d", SizeHeader, n)
 	}
 
-	// Verify request frame fields.
```
soypat, line 32: you deleted several clarifying comments throughout this test, why?

ntp/client_test.go
```diff
@@ -19,7 +42,8 @@ func TestClient_FullExchange(t *testing.T) {
 		t.Fatal("client should not be done before exchange")
 	}
 
-	// Step 1: Client encapsulates request.
+	// --- First exchange ---
```
soypat, line 45: I feel like the original comment was clearer...

ntp/client_test.go
```diff
@@ -45,61 +65,55 @@ func TestClient_FullExchange(t *testing.T) {
...
 	}
-
 	rtd := client.RoundTripDelay()
 	if rtd < 0 {
 		t.Fatalf("round trip delay should be non-negative, got %s", rtd)
 	}
-
-	// Verify Now() returns a reasonable time.
 	ntpNow := client.Now()
 	if ntpNow.Before(baseTime) {
```
soypat, line 118: I'd actually rather you revert this test to original state, comments and everything. observe how it fails (if it fails) and then do small, well directed modifications with comments on why you change each thing. I don't understand why you changed what you changed. ntp was working before these changes.

soypat (changes requested):
Excited to test this out on my systems, but first a couple of questions and suggestions

MDr164:
> Excited to test this out on my systems, but first a couple of questions and suggestions

I think I'm currently working on too many construction sides at once and several items like the removal of some comments slipped under my radar. I was locally experimenting a bit with NTS and inbetween revisions some comments that we're meant to be reintroduced got dropped. The examples are also not done with a lot of effort in mind I need to admit as they, well, just examples. That's also why I used stdlib net in them to make sure we simply test one protocol at a time and for someone implementing a full lneto stack you'd not really copy the example code anyways. But you're right, there are some oversights in the examples and I feel like the scope of this PR so too large for me too keep track of everything. I will split it into several smaller ones for my own sanity. Thanks for the reviews so far!

Btw for testing time related protocols I'm trying out different servers and for NTS specifically either cloudflare or https://time.txryan.com if you also want to test those. I want to get the time related items (NTP, NTS, maybe roughtime and ttt) out of my way before working on VLAN (for which I also got a bit of code done but need to reorganize).

soypat:
idea: Might be worth splitting this into two PRs, ntp improvements and nts?

MDr164:
> idea: Might be worth splitting this into two PRs, ntp improvements and nts?

Yes, I created #101 now to just deal with the NTP side first. I think I captured all your review comments from here in that branch as well. I also explicitly added comments regarding the OriginTime/TransferTime mixup. The RFC is written a bit odd and does not call it server/client but rather machine A/B so you could mix up which one is which

soypat:
It might take me a while to get here. Feel free to rebase this PR so we can visualize changes compared to latest main with NTP changes.

MDr164:
> It might take me a while to get here. Feel free to rebase this PR so we can visualize changes compared to latest main with NTP changes.

No worries, I'll continue with other non-crypto heavy protocols for my current project. I'm still not too sure if I like pulling in a hand-rolled crypto algorithm here as this is supposed to be a networking library. But one would violate the RFC if you do not specify support for this exact algorithm amongst other potential ones. Also the majority of code is the two examples and tests, the actual NTS code is rather slim imo.

MDr164:
Btw did you decide whether or not to bring in the crypto pkg? I'm still torn about having it here. For one it is needed to fulfill the NTS RFC but on the other hand this here isn't lcrypto and handrolled crypto is one thing I learned not to do. The reason I'm still following through with NTS is that a secure foundation of time is the base for TLS certificates to work in the first place and plain NTP could be a risk. As I'm mainly dealing with embedded systems that lack a physical RTC and NTP alone feels too problematic, NTS (or Roughtime) just felt like the missing puzzle piece.

soypat:
Yeah, been coming and going to and from this PR. The main reason I don't merge straight away is because I cannot in good faith review crypto code usage. The way I see it there are a few ways forward:

1. We get a noteworthy crypto expert to review the code
2. Create a parallel repository with security oriented features for lneto. Would be nice to test how easy lneto is to integrate with third party libraries. You can own the repository for the time being until lneto has capability to review crypto oriented code.
3. We create a github organization for lneto ecosystem and you can own the security lneto package, `sneto` if you will. I'll take an active role reviewing what I can review which is the performance, heap allocation patterns and whatnot. 

I support all three. 2 and 3 are more immediately actionable. I am more enthusiastic about 3 though since it gives lneto more weight. [Pion](https://github.com/pion) does 3 and they are doing pretty well. They have plenty of funding coming to them. Let me know what you think Marvin :)

MDr164:
I fully agree that simply merging it like this is probably not the best idea. While option 3 is the most solid moving forward I think I could live with a 2.9 kinda option: I move the crypto code out of this PR, we keep NTS inside lneto/x, document that RFC compliance requires a specific AEAD algorithm (you can also just use stdlib AEAD ones but wouldn't be 100% compliant then) and as a stopgap solution host the crypto code under another org or my private account under a permissive license. The code is reasonably small that I'm comfortable enough hosting it with a disclaimer that it hasn't been audited so one could just import it for NTS. That gives you (and me for that matter) more time to think about a separate org eventually. How does that sound?

soypat:
Sounds like a great plan. I'll give it a review after you extract the crypto specific code then. I'm curious as to how the PR is going to change :thinking:

MDr164:
> I'm curious as to how the PR is going to change

Well I'll drop x/siv fully and update the docs. The NTS code is already made in a way to accept an stdlib crypto.AEAD interface. So all crypto would be external or stdlib. I'd say in a similar way as having HTTPS here next to HTTP but not hosting the TLS code here.

MDr164:
Done, moved out everything that relied on the crypto pkg and the pkg itself. That means I had to comment out a couple tests and drop the example code. The NTS code itself was not affected as mentioned. The pkg now lives at https://github.com/MDr164/aes-siv-cmac until we find a better/longterm solution.

Obviously codecov is now unhappy due to me disabling a bunch of tests.

soypat (approved):

PR #93, opened 2026-04-23
add basic support for connect and send for UDP sockfd's

jwetzell:
Tested this with espradio and got `net.Dial("udp",....)` working.

### DIFF
--- a/x/xnet/stack-berkeley.go
+++ b/x/xnet/stack-berkeley.go
@@ -56,6 +56,7 @@ type StackBerkeley struct {
 	pendingFDs   []socket[pendingSocket]
 	tcpListeners []socket[net.Listener]
 	tcpConns     []socket[net.Conn]
+	udpConns     []socket[net.Conn]
 }
 
 // NewBerkeleyStack wraps the gostack with Berkeley-style calling convention. See [StackBerkeley].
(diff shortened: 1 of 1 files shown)


jwetzell:
Not sure if there really is a reason to have the connections be separate slices. If there might be a need to tell them apart protocol wise like if a function is looking for specific connection by `fd` and expects it to be a tcpConn but it does exist and is actually a udpConn?

x/xnet/stack-berkeley.go
```diff
@@ -114,22 +115,45 @@ func (s *StackBerkeley) Connect(sockfd int, host string, ip netip.AddrPort) erro
...
+			return err
+		}
+		conn, ok := c.(net.Conn)
+		if !ok {
+			return fmt.Errorf("Connect: stack returned non-Conn for protocol %d", pending.sock.protocol)
+		}
+		s.mu.Lock()
+		s.pendingFDs = deleteFD(s.pendingFDs, sockfd)
+		s.udpConns = append(s.udpConns, socket[net.Conn]{sockfd: sockfd, sock: conn})
+		s.mu.Unlock()
```
soypat, line 153: Looks great! Might touch it up a bit since the TCP and UDP code is almost identical, maybe there is a way to reuse the logic between the two implementations

soypat (approved):
Thank you for the contribution!

soypat:
> Not sure if there really is a reason to have the connections be separate slices. If there might be a need to tell them apart protocol wise like if a function is looking for specific connection by fd and expects it to be a tcpConn but it does exist and is actually a udpConn?

I was thinking the same thing. Might give it some thought over the weekend. Feel free to submit another PR!

x/xnet/stack-berkeley.go
jwetzell: That should be pretty doable

PR #94, opened 2026-04-24
cleanup connect function and combine connection slices

jwetzell:
- combine `tcpConns` and `udpConns` into single slice
- reuse logic in Berkeley `Connect`


soypat (approved):
Very nice simplification. Thanks a bazillion <3

PR #102, opened 2026-04-27
Remove trailing whitespaces and add suopprt matrix to README

MDr164:
The support matrix is probably not the most content-full right now but as I only added some obvious missing protocols like the IPv6 variant of DHCP or ICMP but I felt like it might be useful as a general overview. Let me know what you think and I can either change or drop it and only keep the whitespace removal.


soypat (approved):
Thank you for this contribution! Very nice :)

PR #114, opened 2026-06-08
Add RFC 3927 support

### LINKED ISSUE #108: Support RFC 3927 (APIPA)
This is a fairly common thing most stacks tend to have some sort of support for, putting it here mainly to add it to a potential road map. 

- [RFC 3927](https://datatracker.ietf.org/doc/html/rfc3927)
- This is also known as APIPA or self-assigned IPs. Basically those 169.254.x.x a device might give itself in the scenario where no DHCP server is found.

MDr164:
This resolves #108 and adds RFC 3927 support. The code has been written with the support of an LLM (Claude) but tested and validated by me. My current test machine is arm64 based target running tamago (not yet upstream and public). The only item I'm not yet fully contempt with are the comments. They help me understand what the code is doing and are accurate but comments bitrot quite fast from experience and I can't yet tell which section are totally obvious and can be omitted.

### DIFF
--- a/ipv4/linklocal4/handler.go
+++ b/ipv4/linklocal4/handler.go
@@ -1,0 +1,331 @@
+package linklocal4
+
+import (
+	"time"
+
+	"github.com/soypat/lneto"
+	"github.com/soypat/lneto/arp"
+	"github.com/soypat/lneto/ethernet"
+	"github.com/soypat/lneto/internal"
+	"github.com/soypat/lneto/ipv4"
+)
+
+// Handler implements the [RFC3927] IPv4 link-local address autoconfiguration
+// state machine. It is a [lneto.StackNode] over the ARP EtherType: it produces
+// ARP probes and announcements on [Handler.Encapsulate] and inspects incoming
+// ARP traffic for conflicts on [Handler.Demux].
+//
+// Handler is heapless and performs zero allocations after [Handler.Reset];
+// the only allocation is the one-time capture of the clock function. It holds
+// no internal buffers and operates entirely on the caller-supplied scratch
+// buffer, making it suitable for memory constrained targets.
+//
+// A Handler claims and defends a single address. Normal "who-has" ARP
+// resolution of the claimed address is the responsibility of the ARP layer
+// (see [arp.Handler]); this Handler only manages the claim-and-defend protocol.
+//
+// [RFC3927]: https://datatracker.ietf.org/doc/html/rfc3927
+type Handler struct {
+	connID uint64
+	now    func() time.Time
+
+	// nextActionAt is the time at which the next probe/announcement is due.
+	nextActionAt time.Time
+	// lastDefend is the time the most recent defensive announcement was sent.
+	lastDefend time.Time
+
+	prng           uint32
+	candidate      [4]byte
+	firstCandidate [4]byte
+	hw             [6]byte
+
+	state        State
+	probesSent   uint8
+	announceSent uint8
+	conflicts    uint8
+
+	haveFirst   bool
+	defendDue   bool
+	defendValid bool
+	vld         lneto.Validator
+}
+
+var _ lneto.StackNode = (*Handler)(nil)
+
+// linkLocalNet is the RFC3927 IPv4 link-local prefix (169.254.0.0/16). The first
+// and last /24 within it (169.254.0.x and 169.254.255.x) are reserved per
+// section 2.1, so the usable host range is 169.254.1.0-169.254.254.255.
+var linkLocalNet = ipv4.PrefixFrom([4]byte{169, 254, 0, 0}, 16)
+
+// Config configures a [Handler] for link-local address acquisition.
+type Config struct {
+	// HardwareAddr is the interface MAC address used as the ARP sender hardware address.
+	HardwareAddr [6]byte
+	// Now is the monotonic clock source used to schedule probes and announcements.
+	// It is required.
+	Now func() time.Time
+	// Seed seeds the pseudo-random address generator. It is required and must be
+	// non-zero. Per RFC3927 section 2.1 it SHOULD be derived from a persistent
+	// per-host value such as the MAC address so that different hosts pick different
+	// sequences and a host tends to reuse the same address across reboots.
+	Seed uint64
+	// FirstCandidate, if within 169.254.1.0-169.254.254.255, is tried before any
+	// random address. Use it to retry a previously recorded address.
+	FirstCandidate [4]byte
+}
+
+// Reset configures the handler and begins link-local address acquisition,
+// transitioning to [StateWaiting]. It increments the connection ID, invalidating
+// any prior registration.
+func (h *Handler) Reset(cfg Config) error {
+	if cfg.Now == nil || internal.IsZeroed(cfg.HardwareAddr[:]...) || cfg.Seed == 0 {
+		return lneto.ErrInvalidConfig
+	}
+	first := cfg.FirstCandidate
+	haveFirst := linkLocalNet.Contains(first) && first[2] >= 1 && first[2] <= 254
+	*h = Handler{
+		connID:         h.connID + 1,
+		now:            cfg.Now,
+		prng:           uint32(cfg.Seed) ^ uint32(cfg.Seed>>32),
+		hw:             cfg.HardwareAddr,
+		firstCandidate: first,
+		haveFirst:      haveFirst,
+	}
+	if h.prng == 0 {
+		h.prng = 1 // Fold of a non-zero seed can still be zero; xorshift cannot escape the zero state.
+	}
+	h.beginProbing(cfg.Now(), randDelay(h.prand(), probeWait))
+	return nil
+}
+
+// LocalPort implements [lneto.StackNode]. It always returns 0.
+func (h *Handler) LocalPort() uint16 { return 0 }
+
+// Protocol implements [lneto.StackNode], returning the ARP EtherType.
+func (h *Handler) Protocol() uint64 { return uint64(ethernet.TypeARP) }
+
+// ConnectionID implements [lneto.StackNode].
+func (h *Handler) ConnectionID() *uint64 { return &h.connID }
+
+// State returns the current autoconfiguration state.
+func (h *Handler) State() State { return h.state }
+
+// Addr returns the claimed link-local address. ok is true only once the address
+// has been successfully claimed (state [StateBound]).
+func (h *Handler) Addr() (addr [4]byte, ok bool) {
+	return h.candidate, h.state == StateBound
+}
+
+// Candidate returns the address currently being probed, announced or defended.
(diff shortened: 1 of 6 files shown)


soypat (commented):
is `apipa` sufficiently recognizable as a package name instead of `ipv4ll`?

ipv4/linklocal4/handler.go
```diff
@@ -0,0 +1,335 @@
...
+		return 0, nil
+	}
+
+	n, err := h.writeARP(b, senderProto)
+	if err != nil {
+		return 0, err
+	}
+	if offsetToFrame >= 14 {
+		broadcast := ethernet.BroadcastAddr()
+		copy(carrierData[offsetToFrame-14:offsetToFrame-8], broadcast[:])
```
soypat, line 193: Maybe leave a TODO here noting we do not support VLAN tagging here.

ipv4/ipv4ll/handler.go
```diff
@@ -0,0 +1,335 @@
...
+		return
+	}
+	// 254*256 = 65024 usable addresses; offset by one /24 to skip 169.254.0.x.
+	low := 256 + h.prand()%65024
+	h.candidate = [4]byte{169, 254, byte(low >> 8), byte(low)}
+}
+
+// writeARP marshals an ARP request (probe or announcement) into dst. A probe has
+// an all-zero sender protocol address; an announcement repeats the candidate.
+func (h *Handler) writeARP(dst []byte, senderProto [4]byte) (int, error) {
```
soypat: call this putARP or mustWriteARP or something that implies it needs a valid buffer length to write into and have it not return error

ipv4/ipv4ll/handler.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		// onConflict already selected a fresh candidate; resume probing it.
+		h.beginProbing(now, randDelay(h.prand(), probeWait))
+		return 0, nil
+
+	default:
+		return 0, nil
+	}
+
+	n, err := h.writeARP(b, senderProto)
```
soypat: ARP length check before changing state.

ipv4/linklocal4/handler.go
```diff
@@ -0,0 +1,335 @@
...
+
+// Conflicts returns the number of address conflicts encountered so far.
+func (h *Handler) Conflicts() int { return int(h.conflicts) }
+
+// Encapsulate implements [lneto.StackNode]. It writes the next ARP probe or
+// announcement into carrierData at offsetToFrame when one is due, returning the
+// number of bytes written, or 0 when no action is pending. The Ethernet
+// destination, if present before offsetToFrame, is set to broadcast.
+func (h *Handler) Encapsulate(carrierData []byte, _, offsetToFrame int) (int, error) {
+	now := h.now()
```
soypat, line 133: add a length check of buffer early on before changing handler state.

ipv4/ipv4ll/handler.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	afrm, err := arp.NewFrame(carrierData[frameOffset:])
+	if err != nil {
+		return err
+	}
+	var vld lneto.Validator
+	afrm.ValidateSize(&vld)
+	if vld.HasError() {
+		return vld.ErrPop()
+	}
```
soypat: have validator be part of Handler so this does not allocate. also call Reset before validating

ipv4/linklocal4/handler.go
MDr164: I actually am working on more complete VLAN support as my test network here uses VLAN tagging but I'll add it, just forgot.

MDr164:
> is `apipa` sufficiently recognizable as a package name instead of `ipv4ll`?

Not too sure. I was more familiar with IPv4 link-local but maybe APIPA is generally more known? The package name choice was rather opinionated.

ipv4/ipv4ll/handler.go
MDr164: Usually I'm more familiar with the patterns of `doXXX` meaning it can return an error while `mustDoXXX` means it will never return an error but might panic. I'm fine with `putARP` here though.

soypat:
Hmm so after some research it seems microsoft came up with APIPA, which is a point against it. That said I do like `apipa` more than `ipv4ll` which has the unfortunate use of two l's which are not only not as expressive but also easier to confuse with upper case I and 1.

MDr164:
How about just `linklocal`? I already made it a subpkg under ipv4 so the final pkg name would be `ipv4/linklocal`. I guess that's descriptive enough and rather self-explanatory? I have to admit I had to first look up what `APIPA` was after reading the issue title and realized it's just the linklocal stuff I already knew.

soypat:
There is both IPv4 and IPv6 linklocals, so  `linklocal4`? `apipa` is specific to IPv4

MDr164:
`linklocal4` sounds good to me. I just don't associate anything with `apipa` but that is very subjective of course. I'll implement the proposed changes this week together with the rename if that's fine then. :+1:

soypat:
This ready for review?

MDr164:
Yes, I think I captured everything now. I could rebase again but as there are no conflicts right now I suppose there's no value in that.

ipv4/linklocal4/handler.go
```diff
@@ -0,0 +1,333 @@
...
+		connID:         h.connID + 1,
+		now:            cfg.Now,
+		prng:           uint32(seed) ^ uint32(seed>>32),
+		hw:             cfg.HardwareAddr,
+		firstCandidate: first,
+		haveFirst:      haveFirst,
+	}
+	if h.prng == 0 {
+		h.prng = 1 // xorshift cannot escape the zero state.
+	}
```
soypat, line 96: have prng be required non-zero

ipv4/linklocal4/handler.go
```diff
@@ -0,0 +1,333 @@
...
+	if cfg.Now == nil || internal.IsZeroed(cfg.HardwareAddr[:]...) {
+		return lneto.ErrInvalidConfig
+	}
+	seed := cfg.Seed
+	if seed == 0 {
+		// Derive a per-host seed from the MAC address.
+		for _, b := range cfg.HardwareAddr {
+			seed = seed<<8 | uint64(b)
+		}
+	}
```
soypat: seed must be non-zero. I intend on making all seeds in this package non-zero required. I may have missed some though

soypat (changes requested):

ipv4/linklocal4/handler.go
```diff
@@ -0,0 +1,333 @@
...
+	// It is required.
+	Now func() time.Time
+	// Seed seeds the pseudo-random address generator. Per RFC3927 section 2.1 it
+	// SHOULD be derived from a persistent per-host value such as the MAC address so
+	// that different hosts pick different sequences and a host tends to reuse the same
+	// address across reboots. If zero, a seed is derived from HardwareAddr.
+	Seed uint64
+	// FirstCandidate, if non-zero and within 169.254.1.0-169.254.254.255, is tried
+	// before any random address. Use it to retry a previously recorded address.
+	FirstCandidate [4]byte
```
soypat: Also, could this be properly expressed as an ipv4.Prefix? this type has methods that may aid this implementation

ipv4/linklocal4/handler.go
MDr164: Would you expect an error or even panic instead of changing the zero value to one that at least works but provides no good randomization (e.g. like providing 0)? I'm fine with changing it, just want to know what the desired behavior there is. Same question for the seed value I assume though there I think it was mentioned in the RFC to have a fallback in case it was 0.

ipv4/linklocal4/handler.go
MDr164: See comment above regarding expected behavior on zero value

ipv4/linklocal4/handler.go
MDr164: Should be possible, I'll take a look later how I could easily change the API there :)

MDr164: So I changed the API from `[4]byte` to `ipv4.Prefix` but you'll have to call it with e.g. `ipv4.PrefixFrom(xxx, 32)` to get a `/32` length right in this case. It feels only marginally better than a raw byte array tbh but I'll push the API change in a bit

ipv4/linklocal4/handler.go
MDr164: Updated it now

ipv4/linklocal4/handler.go
MDr164: Updated it now

ipv4/linklocal4/handler.go
soypat: I was suggesting ipv4.Prefix because it "felt" like linklocal needs to know the subnet it is on to know local address ranges, after all it is called link local- but I may be absolutely wrong. Does it not help to know the subnet in this case?

ipv4/linklocal4/handler.go
```diff
@@ -0,0 +1,327 @@
...
+// per RFC3927 section 2.1 (the first and last /24 are reserved).
+func (h *Handler) selectCandidate() {
+	if h.haveFirst {
+		h.haveFirst = false
+		h.candidate = h.firstCandidate
+		return
+	}
+	// 254*256 = 65024 usable addresses; offset by one /24 to skip 169.254.0.x.
+	low := 256 + h.prand()%65024
+	h.candidate = [4]byte{169, 254, byte(low >> 8), byte(low)}
```
soypat, line 296: Here! This looks like a place to leverage ipv4.Prefix.Next instead of  assuming /16 subnet, right?

soypat: ah link local has a fixed address range, right? in that case it seems Prefix is overkill

MDr164: Yes, it's fixed. But then I also misunderstood your initial comment above, my bad :)

soypat: No! My bad! Your original code looks to be OK! Link local have a fixed range!

soypat:
I'm sorry about the noise :facepalm: You can revert the Prefix changes if you want. Now I see the fixed-range nature of link local I'm not sure it adds much benefit

MDr164:
No worries, I'll do that later. I'll try to figure if there might be another good usecase of ipv4.Prefix inside the pkg

ipv4/linklocal4/handler.go
MDr164: Updated the code to use the (in my opinion) more ergonomic byte array but used ipv4.Prefix internally to better pronounce that this range is fixed. Hope that's an alright middle ground.

soypat (approved):
Amazing job @MDr164. Thank you for your contribution. Really appreciate these high quality PRs!

PR #77, opened 2026-04-13
TCP Congestion Control

soypat:
Fixes #60. This is a WIP. Might be a while thinking it through

PR #115, opened 2026-06-08
TCP Congestion Control BBR/CUBIC [DO NOT MERGE]

MDr164:
This is a continuation of #77 to not only add the interface but also wire up two different algorithms in order to test, prove and iterate over the proposal. This is pretty experimental and rather hard to properly test on my end, hence I'm keeping this a draft for now. Resolves #60

### DIFF
--- a/tcp/congestion/bbr.go
+++ b/tcp/congestion/bbr.go
@@ -1,0 +1,412 @@
+package congestion
+
+import (
+	"math"
+	"time"
+
+	"github.com/soypat/lneto/tcp"
+)
+
+// BBR tuning constants as defined by BBRv3 [draft-ietf-ccwg-bbr]. BBR
+// ("Bottleneck Bandwidth and Round-trip propagation time") models the network
+// path with two estimated quantities — the maximum delivery rate (BBR.max_bw)
+// and the minimum round-trip time (BBR.min_rtt) — and sizes the volume of
+// in-flight data from the resulting bandwidth-delay product (BDP).
+//
+// [draft-ietf-ccwg-bbr]: https://datatracker.ietf.org/doc/draft-ietf-ccwg-bbr/
+const (
+	// bbrStartupPacingGain ≈ 2.77 = 4*ln(2), the pacing gain that allows the
+	// sending rate to double each round during Startup
+	// (BBR.StartupPacingGain, [draft-ietf-ccwg-bbr] §2.4).
+	bbrStartupPacingGain = 4 * math.Ln2
+	// bbrDrainPacingGain is the pacing gain used in Drain to empty the queue
+	// built during Startup within one round trip: any value at or below
+	// 1/BBR.DefaultCwndGain = 0.5 suffices and BBR uses 0.5
+	// (BBR.DrainPacingGain, [draft-ietf-ccwg-bbr] §2.4, §5.3.2).
+	bbrDrainPacingGain = 0.5
+	// bbrDefaultCwndGain is the cwnd gain used in most phases (Startup, Drain,
+	// ProbeBW): scaling the BDP by 2 allows the sending rate to double each
+	// round and leaves headroom for delayed/aggregated ACKs
+	// (BBR.DefaultCwndGain, [draft-ietf-ccwg-bbr] §2.5).
+	bbrDefaultCwndGain = 2.0
+	// bbrFullBwThresh is the minimum per-round delivery-rate growth ratio that
+	// still counts as "the pipe is still filling" during Startup: less than 25%
+	// growth counts toward Startup exit ([draft-ietf-ccwg-bbr] §5.3.1.2).
+	bbrFullBwThresh = 1.25
+	// bbrFullBwCount is the number of consecutive non-app-limited rounds
+	// without significant bandwidth growth required to estimate the pipe is
+	// full and exit Startup ([draft-ietf-ccwg-bbr] §5.3.1.2).
+	bbrFullBwCount = 3
+	// bbrMinPipeCwnd is the minimal congestion window BBR targets, in segments,
+	// allowing pipelining with delayed-ACK peers (BBR.MinPipeCwnd = 4*SMSS,
+	// [draft-ietf-ccwg-bbr] §2.7).
+	bbrMinPipeCwnd = 4
+	// bbrProbeRTTCwndGain scales the BDP to produce the congestion window held
+	// during ProbeRTT, reducing in-flight data to 50% of the estimated BDP
+	// (BBR.ProbeRTTCwndGain, [draft-ietf-ccwg-bbr] §2.13.2).
+	bbrProbeRTTCwndGain = 0.5
+	// bbrProbeRTTDuration is the minimum duration for which ProbeRTT holds the
+	// reduced in-flight volume to drain the path and re-measure min_rtt
+	// (BBR.ProbeRTTDuration, [draft-ietf-ccwg-bbr] §2.13.2).
+	bbrProbeRTTDuration = 200 * time.Millisecond
+	// bbrProbeRTTInterval is the minimum time interval between ProbeRTT states:
+	// a min_rtt sample older than this schedules a ProbeRTT
+	// (BBR.ProbeRTTInterval, [draft-ietf-ccwg-bbr] §2.13.2). This simplified
+	// implementation uses it directly as the min_rtt staleness window instead
+	// of keeping the separate 10-second BBR.MinRTTFilterLen of §2.13.1.
+	bbrProbeRTTInterval = 5 * time.Second
+	// bbrBwWindowRounds is the length, in round trips, of the max_bw max
+	// filter. It approximates the BBR.MaxBwFilterLen window of 2 ProbeBW
+	// cycles ([draft-ietf-ccwg-bbr] §2.10) for this implementation's
+	// fixed-duration gain cycling, where a full cycle spans roughly
+	// len(bbrPacingGainCycle) round trips.
+	bbrBwWindowRounds uint32 = 2 * uint32(len(bbrPacingGainCycle))
+)
+
+// bbrPacingGainCycle is the pacing-gain sequence used in ProbeBW, a
+// fixed-duration simplification of the ProbeBW_UP/DOWN/CRUISE/REFILL phases of
+// [draft-ietf-ccwg-bbr] §5.3.3: the 1.25 phase probes for more bandwidth
+// (ProbeBW_UP, §5.3.3.4), the 0.9 phase drains any queue the probe created
+// (ProbeBW_DOWN, §5.3.3.1) and the unity phases cruise at the estimated
+// bandwidth (ProbeBW_CRUISE/REFILL, §5.3.3.2, §5.3.3.3). Each phase lasts
+// about one min_rtt rather than using the draft's adaptive phase durations.
+var bbrPacingGainCycle = [8]float64{1.25, 0.9, 1, 1, 1, 1, 1, 1}
+
+// bbrState enumerates the BBR state machine phases.
+type bbrState uint8
+
+const (
+	bbrStartup  bbrState = iota // Exponentially probe for bandwidth.
+	bbrDrain                    // Drain the queue created during Startup.
+	bbrProbeBW                  // Steady-state: cycle pacing gain around 1.0.
+	bbrProbeRTT                 // Periodically drain to re-measure min_rtt.
+)
+
+func (s bbrState) String() string {
+	switch s {
+	case bbrStartup:
+		return "STARTUP"
+	case bbrDrain:
+		return "DRAIN"
+	case bbrProbeBW:
+		return "PROBE_BW"
+	case bbrProbeRTT:
+		return "PROBE_RTT"
+	default:
+		return "<invalid bbr state>"
+	}
+}
+
+// BBRConfig configures a [BBR] controller. See [BBR.Reset].
+type BBRConfig struct {
+	// MSS is the maximum segment size in bytes. If zero, a default of 1460 is used.
+	MSS tcp.Size
+	// InitialCwnd is the initial congestion window in segments. If zero, the
+	// [RFC6928] recommended value of 10 segments is used.
+	InitialCwnd tcp.Size
+	// Now injects a clock for deterministic testing. If nil, [time.Now] is used.
+	Now func() time.Time
+}
+
+// BBR implements a simplified version of the BBRv3 congestion-control
+// algorithm specified in [draft-ietf-ccwg-bbr]. Rather than reacting to loss
+// like Reno/CUBIC, BBR continuously estimates the maximum delivery rate
+// (max_bw) and minimum round-trip time (min_rtt) and sizes the congestion
+// window from the resulting bandwidth-delay product, probing periodically for
+// changes. It implements tcp.CongestionControl.
+//
+// Simplifications relative to the draft: round trips are approximated by
+// elapsed time instead of packet-delivery accounting (§5.5.1); the ProbeBW
+// phases use fixed durations of one min_rtt with the draft's gain values
+// instead of adaptive phase lengths (§5.3.3); the loss-based short-term model
+// (BBR.Beta, BBR.LossThresh of §2.7) and the extra_acked aggregation estimator
+// (§5.5.9) are not implemented.
+type BBR struct {
+	base ccBase
+
+	state bbrState
+
+	bwFilter    minmax // windowed max of delivery rate, bytes/sec.
+	roundCount  uint32
+	roundStart  time.Time
+	roundInited bool
+
+	rtProp      time.Duration
+	rtPropStamp time.Time
+
+	pacingGain float64
+	cwndGain   float64
+	cwnd       tcp.Size // congestion window, bytes.
+
+	cycleIndex int       // current phase within bbrPacingGainCycle.
+	cycleStamp time.Time // when the current ProbeBW phase began.
+
+	fullBw        float64 // bandwidth at the last full-pipe check, bytes/sec.
+	fullBwCount   int
+	fullBwReached bool
+
+	probeRTTDone     time.Time
+	probeRTTDoneInit bool
+}
+
+// Reset (re)initializes the controller with cfg, returning an error if the
+// configuration is invalid. It follows the package Reset/Configure convention.
+func (bbr *BBR) Reset(cfg BBRConfig) error {
+	mss := cfg.MSS
+	if mss == 0 {
+		mss = defaultMSS
+	}
+	icwnd := cfg.InitialCwnd
+	if icwnd == 0 {
+		icwnd = 10
+	}
+	*bbr = BBR{
+		base: ccBase{
+			mss:   cfg.MSS,
+			clock: cfg.Now,
+		},
+		state:      bbrStartup,
+		pacingGain: bbrStartupPacingGain,
+		cwndGain:   bbrDefaultCwndGain,
+		cwnd:       icwnd * mss,
+	}
+	return nil
+}
+
+// State returns the current BBR state-machine phase as a human-readable string
+// ("STARTUP", "DRAIN", "PROBE_BW" or "PROBE_RTT").
+func (bbr *BBR) State() string { return bbr.state.String() }
+
+// CongestionWindow returns the current congestion window in bytes.
+func (bbr *BBR) CongestionWindow() tcp.Size { return bbr.cwnd }
+
+// BandwidthEstimate returns the current maximum delivery rate estimate (max_bw)
+// in bytes per second, or 0 before the first delivery-rate sample.
+func (bbr *BBR) BandwidthEstimate() float64 { return float64(bbr.bwFilter.get()) }
+
+// MinRTT returns the current minimum round-trip time estimate (min_rtt), or 0
+// before the first RTT sample.
+func (bbr *BBR) MinRTT() time.Duration { return bbr.rtProp }
+
+// PacingRate returns the rate, in bytes per second, at which the sender should
+// pace transmissions: pacing_gain * max_bw.
+func (bbr *BBR) PacingRate() float64 { return bbr.pacingGain * bbr.BandwidthEstimate() }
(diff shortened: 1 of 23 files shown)


tcp/definitions.go
```diff
@@ -10,6 +10,63 @@ import (
...
+}
+
+// CongestionEvent describes a single TCP segment crossing the connection,
+// together with the sender state needed by a [CongestionControl] to update its
+// estimates. The Handler builds one of these for every segment it sends or
+// receives and passes it to [CongestionControl.Control].
+type CongestionEvent struct {
+	// SEQ and ACK are the segment's sequence and acknowledgment numbers.
+	SEQ Value
+	ACK Value
```
soypat: Seems like we can group some fields here into a `Incoming Segment` field which will make it more self-documenting

tcp/congestion/cubic.go
```diff
@@ -0,0 +1,245 @@
...
+func (c *CUBIC) cubicTarget(t float64) float64 {
+	d := t - c.k
+	return c.originPoint + cubicC*d*d*d
+}
+
+// OnLoss applies a multiplicative window decrease in response to a congestion
+// event (e.g. three duplicate ACKs). Repeated calls within the same loss epoch
+// are coalesced into a single reduction, matching TCP's "one cut per window of
+// data" behaviour.
+func (c *CUBIC) OnLoss() {
```
soypat: rename (cubic *CUBIC)

tcp/handler.go
```diff
@@ -28,13 +28,44 @@ type Handler struct {
...
 	nRetransmit uint8
 }
 
+// SetCongestionControl installs cc as the connection's congestion controller,
+// or removes it when cc is nil. It limits how much new (unacknowledged) data
+// the Handler keeps in flight to cc.CongestionWindow bytes and feeds every
+// segment crossing the connection into cc.Control. The controller is retained
+// across connection re-opens. It is typically configured before the connection
+// is opened (see [ConnConfig.CongestionControl]).
+func (h *Handler) SetCongestionControl(cc CongestionControl) { h.cc = cc }
```
soypat: Is changing congestion control mid-connection something we want to allow? I'd keep it limited to Configure. A sign someone is changing the congestion algorithm mid-connection seems like a sign of a leaky abstraction, maybe means a method is missing on the interface.

tcp/handler.go
```diff
@@ -218,6 +250,11 @@ func (h *Handler) Recv(incomingPacket []byte) error {
 			h.remotePort = remotePort
 		}
 	}
+	if h.cc != nil {
+		// Feed the received segment (ACKs, duplicate ACKs/loss, RTT samples) into
+		// the congestion controller after the TCB has updated snd.UNA/dupack.
+		h.cc.Control(h.congestionEvent(segIncoming, false))
```
soypat: A single method interface may simplify API greatly.
```go
const invalidCongestWnd = 0xffff_ffff
var congestWnd Size = invalidCongestWnd
if h.cc != nil {
    congestWnd = h.cc.Control(...)
}
...
    if congestWnd != invalidCongestWnd && !h.scb.HasPendingRetransmit() {
        ...
        if congestWnd > inflight {
				avail = congestWnd - inflight
		}
    }
```

tcp/definitions.go
```diff
@@ -10,6 +10,63 @@ import (
...
+// implementations: congestion.CUBIC and congestion.BBR.
+//
+// The interface is intentionally decoupled from the [Handler]'s internals: the
+// Handler hands the controller a fully-populated [CongestionEvent], so
+// implementations live in a separate package and never touch unexported state.
+type CongestionControl interface {
+	// Control feeds a segment crossing the [Handler] boundary into the
+	// controller so it can update its congestion state. It is called for every
+	// admitted segment, not only on congestion events.
+	Control(event CongestionEvent)
```
soypat: I do like your API more than the one I proposed based on signature solely. My only worry is that it may give users less functionality than the one I had previously proposed:

```go
func (Congestion) Control(h *Handler, segment Segment, isTx bool) (retransmit []int, haltSendingNewPackets bool)
```

Are we sure this covers all congestion use cases users will want in the future? I'm pretty much 99% sold on your API, but I don't want to not give it due thought and consideration it deserves. We can always change it in the future fyi.

tcp/definitions.go
```diff
@@ -10,6 +10,63 @@ import (
...
+	// MSS is the maximum segment size negotiated with the peer, or 0 if unknown.
+	MSS Size
+	// Flags are the segment's TCP flags.
+	Flags Flags
+	// Tx reports whether the segment was transmitted by this endpoint (true) or
+	// received from the peer (false).
+	Tx bool
+	// Loss reports a congestion/loss signal for this event, such as the
+	// duplicate-ACK retransmission threshold having been reached.
+	Loss bool
```
soypat: Including dupack number `uint8` maybe gives information with same struct size? Not sure what other numbers measure loss- and not really sure if it is more valuable than a simple bool.

tcp/handler.go
```diff
@@ -28,13 +28,44 @@ type Handler struct {
...
+		SndNXT:   h.scb.snd.NXT,
+		DataLen:  seg.DATALEN,
+		InFlight: h.scb.snd.inFlight(),
+		MSS:      h.scb.snd.MSS,
+		Flags:    seg.Flags,
+		Tx:       tx,
+		Loss:     !tx && h.scb.dupack >= retransmitAfterDupacks,
+	}
+}
+
```
soypat: this looks more like a ControlBlock method than a Handler method, no? If added to ControlBlock have it be exported

soypat (commented):
a few questions and comments about interface before I read the actual congestion package implementations. Really happy someone took this task on! Thank you Marvin!

MDr164:
> Really happy someone took this task on!

I'll slowly continue updating this PR but it'll probably take a couple more iterations until we can finalize what interface we want to actually go for. I opted for CUBIC and BBR specifically as CUBIC is the defacto default under Linux and a loss based algorithm while BBR is a more modern and promising model based algorithm. Those two should be a good field study on how to design our interface right. I'm still trying to create a decent enough testing setup for this though to put actual numbers and not just speculation into the PR.

tcp/definitions.go
MDr164: > Are we sure this covers all congestion use cases users will want in the future?

No, if I could confidently predict things like that I'd go gambling :D

Right now retransmission control stays within the congestion handler. We could grow the CongestionEvent struct could also grow more fields so we would not need to modify the Control() function signature in the future.

tcp/definitions.go
MDr164: Done, named it just `Segment Segment` instead of `Incoming Segment`

tcp/congestion/cubic.go
MDr164: Done

tcp/handler.go
MDr164: Probably not, you're right. Changed it to returning an error instead of allowing reconfiguration while connected. Probably saves a lot of headaches in the future.

tcp/handler.go
MDr164: Done (in a similar way)

tcp/definitions.go
MDr164: Done

tcp/handler.go
MDr164: True. It kind of grew organically and I didn't fully plan out where to place what yet and just wanted a basic poc running. Moved it over now.

tcp/handler.go
```diff
@@ -28,13 +28,40 @@ type Handler struct {
...
+// SetCongestionControl installs cc as the connection's congestion controller,
+// or removes it when cc is nil. It limits how much new (unacknowledged) data
+// the Handler keeps in flight to the window returned by cc.Control, which is
+// fed every segment crossing the connection. The controller is retained across
+// connection re-opens. Returns an error if the connection is open: the
+// controller cannot be changed mid-connection (see
+// [ConnConfig.CongestionControl] to configure it on a [Conn]).
+func (h *Handler) SetCongestionControl(cc CongestionControl) error {
+	if !h.scb.State().IsClosed() {
+		return lneto.ErrInvalidConfig
```
soypat: This is spiritually the same error as `errNeedClosedTCBToOpen`. Maybe worth adding a new generic error to lneto and share it unless there's a better error than ErrInvalidConfig, because ErrInvalidConfig signals the cc variable is invalid, which cannot happen.

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,412 @@
...
+	rtPropStamp time.Time
+
+	pacingGain float64
+	cwndGain   float64
+	cwnd       tcp.Size // congestion window, bytes.
+
+	cycleIndex int       // current phase within bbrPacingGainCycle.
+	cycleStamp time.Time // when the current ProbeBW phase began.
+
+	fullBw        float64 // bandwidth at the last full-pipe check, bytes/sec.
```
soypat: Are the gains from a float64 worth it over a float32?

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,412 @@
...
+	fullBwCount   int
+	fullBwReached bool
+
+	probeRTTDone     time.Time
+	probeRTTDoneInit bool
+}
+
+// Reset (re)initializes the controller with cfg, returning an error if the
+// configuration is invalid. It follows the package Reset/Configure convention.
+func (bbr *BBR) Reset(cfg BBRConfig) error {
```
soypat: I believe we may need to have a `func (*BBR) Reset() method signature and a separate Configure method... Here's what I suspect would be better:

- `Control(CongestionEvent)`: Send/receive and return congestion window, the actual functionality
- `Reset()`: Reset the congestion algorithm state so it is ready for a new connection. Called when Handler first opens a connection. Is likely part of tcp.CongestionControl interface.
- `Configure(XConfig) error`: configuration of the algorithm before being set on a Handler. Not part of tcp.CongestionControl algorithm

Let me know what you think!

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,412 @@
...
+
+// PacingRate returns the rate, in bytes per second, at which the sender should
+// pace transmissions: pacing_gain * max_bw.
+func (bbr *BBR) PacingRate() float64 { return bbr.pacingGain * bbr.BandwidthEstimate() }
+
+// BDP returns the bandwidth-delay product in bytes: max_bw * min_rtt. It is the
+// amount of in-flight data needed to keep the bottleneck fully utilized.
+func (bbr *BBR) BDP() float64 {
+	return bbr.BandwidthEstimate() * bbr.rtProp.Seconds()
+}
```
soypat, line 227: I'm curious why you want to export methods other than the ones needed for use

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,412 @@
...
+	bbr.cwnd = target
+}
+
+// OnLoss is provided for symmetry with [CUBIC]. BBR is not loss-based: it does
+// not collapse its window on packet loss the way Reno/CUBIC do, so this is
+// intentionally a no-op. The draft's loss-driven short-term model
+// (BBR.Beta = 0.7 and BBR.LossThresh = 2%, [draft-ietf-ccwg-bbr] §2.7,
+// §5.5.10) is not implemented; bandwidth and RTT estimation alone drive the
+// window via OnACK.
+func (bbr *BBR) OnLoss() {}
```
soypat: whats the plan here?

MDr164: As mentioned in the comment above the method, it's simply there for symmetry with the other algorithm. No real plan. After all it's a draft to discuss an API using two examples :)

tcp/congestion/bbr.go
MDr164: As I'm mostly just exploring how the API could look like I simply exported rather more than less to explore if they could be used somehow. There is currently no real technical reason, just exploring API :sweat_smile:

tcp/congestion/bbr.go
MDr164: Probably not, I'm right now testing lneto code using 1G and 10G physical hardware so there's plenty breathing room. I could probably add some analysis if there are other variables using overly large data types right now.

soypat: OK, if float32 is equivalent lets use that so its nicer to embedded architectures!

tcp/congestion/bbr.go
MDr164: I have no objections, I'm not fully sure if there are usecases in which one would need to call `Reset()` other than on opening a new connection but decoupling it from `Configure()` doesn't hurt.

soypat: a connection should be reusable without calling configure, merely relying on Abort/Close and Dial/Listen semantics. I imagine the CongestionControl(CC) needs to be notified the connection was tore down/aborted/reset and needs to start anew to work correctly, right? If CC is stateless then disregard this, a single method interface should be OK to represent it.

soypat:
Take your time with this. I'll let you know I'm eagerly waiting for it to be Ready for Review :)

Have not taken a look at tests yet but consider behavioural interface tests. What I mean is have a test that does roughly:

```go
var tests []struct{ CC tcp.CongestionControl }{
    {CC: new(BBR)},
    {CC: new(Cubic)},
}
for _, test := range tests {
    cc := test.CC
    name := reflect.TypeOf(cc).String() // ask AI to do this correctly and omit package name, dot and pointer star.
    t.Run("stateless "+name, func(t *testing.T) { ... })
    t.Run("congest-on-drop "+name, func(t *testing.T) { ... })
    t.Run("no-congest-on-ok "+name, func(t *testing.T) { ... })
}
```

MDr164:
> I'll let you know I'm eagerly waiting for it to be Ready for Review :)

I'll move it out from the draft state once I think it's fine, which right now I just can't warrant with the little testing I did on actual hardware. We got a small hackathon planned with our department here and I plan on working on lneto/tamago then, I think doing actual congestion testing will be one of the first things I'll try and polish up.

> Have not taken a look at tests yet but consider behavioural interface tests. What I mean is have a test that does roughly:
> 
> ```go
> var tests []struct{ CC tcp.CongestionControl }{
>     {CC: new(BBR)},
>     {CC: new(Cubic)},
> }
> for _, test := range tests {
>     cc := test.CC
>     name := reflect.TypeOf(cc).String() // ask AI to do this correctly and omit package name, dot and pointer star.
>     t.Run("stateless "+name, func(t *testing.T) { ... })
>     t.Run("congest-on-drop "+name, func(t *testing.T) { ... })
>     t.Run("no-congest-on-ok "+name, func(t *testing.T) { ... })
> }
> ```

I'd even go as far and consider adding some randomization into the congestion patterns (with a seed to be able to repro locally). I think for now it's rather important to focus on the actual congestion API and not make sure the algorithms themselves are perfect. I'll ponder a bit if there are decently good ways to test and will also take a look at how Linux implements it again.

MDr164:
I added some link emulation testing and uncovered some issues already and described them in the commit messages as well as the code/test comments. Namely:  ControlBlock accepts only in-order segments (no out-of-order buffering), has no RTO timer in the Handler, recovers loss only via 3-dup-ACK fast retransmit, and the receiver aborts after 8 consecutive out-of-order segments. A single mid-stream loss therefore degenerates into go-back-N that deadlocks once the sender runs out of new data to elicit dup-ACKs. Until those are fixed I'd rather keep this a draft.

MDr164:
Ok with #130 merged I got some ideas on how to update this PR now.

MDr164:
The scope exploded a bit again but I'm now happy with the contents of this branch. Though I think now it's time to split this into smaller PRs. I already went ahead to split it into smaller commits.

In summary: TCP congestion control was a can of worms I opened as the actual link testing revealed some shortcomings to make this actually work. This can probably be turned into 5-ish PRs

MDr164:
Ready to review BUT I would actually turn this into multiple PRs instead of one gigantic one that adds several features at once. Feel free to look at individual commits while I ponder how to split this into separate topics.

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,440 @@
...
+
+// BBRConfig configures a [BBR] controller. See [BBR.Configure].
+type BBRConfig struct {
+	// MSS is the maximum segment size in bytes. If zero, a default of 1460 is used.
+	MSS tcp.Size
+	// InitialCwnd is the initial congestion window in segments. If zero, the
+	// [RFC6928] recommended value of 10 segments is used.
+	InitialCwnd tcp.Size
+	// Now injects a clock for deterministic testing. If nil, [time.Now] is used.
+	Now func() time.Time
```
soypat, line 108: We use Nanotime in the rest of this repo: `func () int64` is the signature

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,440 @@
...
+func (bbr *BBR) onACK(acked, inflight tcp.Size, rtt time.Duration) {
+	if acked == 0 || rtt <= 0 {
+		return
+	}
+	bbr.update(float64(acked)/rtt.Seconds(), inflight, rtt)
+}
+
+// update advances the BBR model with a delivery-rate sample of rate bytes/sec
+// (0 means no rate could be measured yet) and a round-trip sample rtt.
+func (bbr *BBR) update(rate float64, inflight tcp.Size, rtt time.Duration) {
```
soypat, line 267: float32

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,440 @@
...
+func (bbr *BBR) MinRTT() time.Duration { return bbr.rtProp }
+
+// pacingRate returns the rate, in bytes per second, at which the sender should
+// pace transmissions: pacing_gain * max_bw.
+func (bbr *BBR) pacingRate() float32 { return bbr.pacingGain * bbr.BandwidthEstimate() }
+
+// bdp returns the bandwidth-delay product in bytes: max_bw * min_rtt. It is the
+// amount of in-flight data needed to keep the bottleneck fully utilized.
+func (bbr *BBR) bdp() float32 {
+	return bbr.BandwidthEstimate() * float32(bbr.rtProp.Seconds())
```
soypat, line 226: maybe worth adding `durationf32()` function to convert time.Duration to float32

tcp/congestion/bbr.go
```diff
@@ -0,0 +1,440 @@
...
+)
+
+// BBR tuning constants as defined by BBRv3 [draft-ietf-ccwg-bbr]. BBR
+// ("Bottleneck Bandwidth and Round-trip propagation time") models the network
+// path with two estimated quantities — the maximum delivery rate (BBR.max_bw)
+// and the minimum round-trip time (BBR.min_rtt) — and sizes the volume of
+// in-flight data from the resulting bandwidth-delay product (BDP).
+//
+// [draft-ietf-ccwg-bbr]: https://datatracker.ietf.org/doc/draft-ietf-ccwg-bbr/
+const (
```
soypat, line 17: I think we should choose a single implementation (cubic?) for the PR and see how that flies. Once the API looks to be established and good we can add more implementations

tcp/reassembly.go
```diff
@@ -0,0 +1,173 @@
...
+// sequence number so that, once the gap is filled by a retransmission, the
+// buffered tail can be delivered without the peer re-sending it (avoiding
+// go-back-N). It is the receiver-side state a SACK sender relies on.
+//
+// Storage is a user-provided slab divided into a fixed number of equal slots,
+// one per held segment, so memory is bounded and nothing is allocated on the
+// data path — suited to the embedded targets lneto serves. A segment larger
+// than a slot, a duplicate, or one that does not fit is simply not stored; the
+// sender retransmits it (and, as a backstop, the RFC 6298 timer recovers it).
+type reassembly struct {
```
soypat, line 16: I'm thinking we can maybe join txqueue and this reassembly logic in one generic block buffer at a low level. They are virtually the same concept.

soypat (changes requested):
I got halfway through according to github scroll, though likely not even that, and pretty much am thinking this is too large to review in one swoop. Large in loc and in scope. 

I'm also thinking I'd rather keep most of the logic you've added to the tcp package as external to the Conn/Handler/ControlBlock types. I'm 99% sure we can think of an interface that captures the timing/congestion/packetloss abstraction in a few simple method signatures which it itself contains congestion-control/RTO/SACK. I feel the path forward might be best and most directly solved with solving **reassembly first** and consolidating txqueue+rx reassembly abstraction at a low level. Once that is solved, suppose we have a tcp.BlockBuffer/tcp.SeqBuffer/whateverBuffer abstraction-  we can expose the underlying buffers on the handler and let the interface operate on them to do SACK/CongestionControl/RTO/etc without having to include all this logic inside the tcp package. That would be a huge win in maintainability/testability of tcp package and would offer tons of flexibility to users wanting to theorycraft advanced TCP algorithms.

Or maybe I'm wrong and we eventually tear down tcp package compactness. Still kinda wanna think about alternative interfaces and consolidating txqueue+rxreassembly.

soypat:
Also, since it seems this requires more thought I'll move it to v0.3.0 milestone

MDr164:
> Still kinda wanna think about alternative interfaces and consolidating txqueue+rxreassembly.

Yes, I got most of Friday open to work on just this item and will spend some time to properly bisect everything. I also did not anticipate this to inflate so much. I think none of this is needed for 0.2.0 as it's purely about performance and reliability improvements while basic TCP has been working for quite some time so we can push it out for sure. I think I'll roughly split this PR into following smaller PRs (loosely ressembling the commits in here as well):
- Keeping this PR for the general overview/bigger picture
- One to add the new interfaces only
- One to add the RFC 6298 items
- One to add the SACK stuff
- One for the RFC 7323 items
- One for the ooo handling
- One for actually adding in CUBIC and BBR

MDr164:
Fret not, I'm still working on this. I think I got a decent split now with 8 branches all building on top of each other (for the most and two of them being less than 40 lines of code change). I'll keep this one here open for the time being and open the other PRs. I apologize for the "flood" of PRs stemming from this :D

MDr164:
Moved back to draft state due to splitting it into multiple smaller PRs

Issue #117, opened 2026-06-08
CI: catch `//go:generate` diff


soypat:
We have no CI pipeline to catch failing to run `go generate ./...` on new changes to enums and whatnot. Problem in action: https://github.com/soypat/lneto/pull/116 ran `go generate` and created a new file and edited a few files that should have been edited from previous `go generate`

soypat:
After implementing it I figured it was a potential vulnerability in CI. will not implement.

PR #120, opened 2026-06-09
add go generate to CI to solve #117

soypat:


soypat:
This is a actually a potential vulnerability in CI. excluding.

PR #122, opened 2026-06-12
docs: update README to aid Linux debugging

### LINKED ISSUE #26: Recommend disabling receive offloading when testing under Linux
On my Linux laptop, all network interfaces run with generic-receive-offload enabled (at least). This results in the kernel merging TCP frames together and delivering them to the raw socket as one larger frame. This is not an issue by itself, but it looks like, at least on my machine, the kernel doesn't bother to recalculate the full checksum for these new frames (it calculates only the pseudo-header one), causing lneto to drop them.
Once I figured this out, which took a while, the solution was pretty simple:

`sudo ethtool -K <network-interface> gro off`

It may be necessary to disable other offloading features in some cases, but for me this was enough for both the wired and wireless interfaces I'm using.

I think there should be some info about this somewhere, perhaps in the README.

MDr164:
This resolves #26 and informs people how to better debug under Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,14 @@ go mod download github.com/soypat/lneto@latest
 
 ## Developing (linux)
 
+When testing lneto through a Linux network interface, disable receive offloading if packets are unexpectedly dropped with checksum errors. Generic receive offload (GRO) can merge TCP frames before delivery to raw sockets without recalculating the full checksum.
+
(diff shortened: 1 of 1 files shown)


soypat (approved):

soypat:
Thank you @MDr164

PR #123, opened 2026-06-15
More IPv6 support and race condition fixes

soypat:


x/xnet/stack-go.go
```diff
@@ -67,31 +67,35 @@ func (s StackGo) Socket(ctx context.Context, network string, family, sotype int,
...
-			// Outbound (dial) connection: auto-assign ephemeral port.
-			laddr = netip.AddrPortFrom(laddr.Addr(), uint16(49152+s.blk.async.Prand32()%16384))
+		// Auto-assign an ephemeral port for both outbound dials and for listeners
+		// that did not request a fixed port.
+		laddr = netip.AddrPortFrom(laddr.Addr(), uint16(49152+s.blk.async.Prand32()%16384))
+	}
+	if laddr.Addr().IsUnspecified() {
+		// Fill in the stack's configured address for the requested family.
+		if isV6 {
+			laddr = netip.AddrPortFrom(netip.AddrFrom16(s.blk.async.Addr6()), laddr.Port())
```
MDr164, line 95: If I get that right: when isV6 and IPv6 is disabled, Addr6() returns all-zeros which is unspecified, then RegisterListenerUDP6/TCP6 returns ErrUnsupported. That's a correct end state, but the failure surfaces late. A !ipv6enabled early-return mirroring the AF check would be clearer imo. The Register*6 dial/register methods do guard properly but feels a bit odd to me.

PR #126, opened 2026-06-18
add zero os docs

soypat:


README.md
```diff
@@ -11,11 +11,14 @@ Userspace networking primitives.
 
 ## Features
 `lneto` provides the following features:
+- Zero Operating System required. ***Zero***. All protocols (including TCP, excepting NTP) are time-independent.
```
MDr164, line 14: I'm currently in the means of adding optional time integration in TCP to better deal with the congestion control items. I deliberately kept it optional so seems like that was a good call.

README.md
```diff
@@ -11,11 +11,14 @@ Userspace networking primitives.
...
+    - Explicit HAL needed. Lneto performs no `time.Sleep` nor `time.Now` calls unless configured by user.
+    - Zero scheduling required. No goroutines/channels use in Lneto. Can be run in event loop.
```
MDr164, line 16: I actually thought about optionally allowing for usage of channels for the io_uring-style changes eventually planned. I suppose I'll try to keep those optional as well so that DCE can make sure they don't land in the binary when compiled without.

MDr164 (approved):

README.md
soypat: Yes! Optional is what we want for the time being :)

README.md
soypat: Yes! I think we might end up developing parallel stack implementations i.e: internet.StackPortsIouring or something similar

soypat:
Thanks for the review!

PR #127, opened 2026-06-18
Ack dropping - fix for #50

### LINKED ISSUE #50: ACKs silently dropped in FIN-WAIT-2
Had a program that calls tcp.Conn.Close without sending a HTTP header with Connection:close resulting in a slowdown. 

The visible slowdown is caused primarily by a keep-alive violation: the server replies with HTTP/1.1 then immediately Close()s without sending Connection: close, so the browser pipelines a second request on the same connection which is never processed (→ ~1.1s timeout/retry). 

Meanwhile, the issue within lneto is a strict out-of-order rejection and ACK behavior in FIN-WAIT-2 are secondary robustness issues but not the root cause of the observed latency, maybe worth a fix. The out of order aspect is out of scope for now, too complex.

AI Summary: 

<details>

1. Out-of-order segment rejection (seq != rcv.nxt)

lneto currently requires segments to arrive in exact sequence order. At 43.379 a valid ACK with seq=0xa75b6b66 arrived before its data (seq=0xa75b6960) — the router delivered them out of order. lneto rejected it and sent a challenge ACK.

The fix is an out-of-order receive queue. When a segment arrives with seg.seq > rcv.nxt, instead of rejecting, buffer it and send a duplicate ACK (signaling the gap). When the missing segment arrives and fills the gap, deliver them all in order. This is standard TCP behavior (RFC 5681 §3.2). The queue can be small — even 2–4 slots would handle most real-world reordering.

Without the full queue, a cheaper partial fix: at minimum, don't reject pure ACKs (data=0) that are ahead of rcv.nxt. The seq field on a pure ACK is informational — what matters is the ack field. The current code rejects it as out-of-order when it should just process the ACK number and ignore the seq.

2. Data received in FIN-WAIT-2 is ACKed but silently dropped

At 43.905, 512 bytes of a new HTTP request arrive on a connection in FIN-WAIT-2. lneto ACKs the data (telling the remote "I got it") but nobody reads it. The remote thinks the request was received.

This is technically RFC-compliant — FIN means "I'm done sending", not "I'm done receiving." But the application already called Close(), so the data goes nowhere.

Two options:

Option A (simple): When the application has called Close() and the connection enters FIN-WAIT-*, send RST instead of ACK for any incoming data segments. This immediately tells the remote "this data was not processed" so it can retry on a new connection. This is what Linux does with SO_LINGER=0.
Option B (correct but complex): Implement Shutdown(WR) vs Close() properly — Close() means both halves are done, so RST on new data. Shutdown(WR) sends FIN but keeps reading. This is the full POSIX model.
Option A is probably the right tradeoff for a microcontroller stack. The user-visible effect: browsers detect the failed request instantly (RST) instead of waiting for a timeout.

</details>

soypat:
Fixes #50

### DIFF
--- a/tcp/handler.go
+++ b/tcp/handler.go
@@ -186,6 +186,14 @@ func (h *Handler) Recv(incomingPacket []byte) error {
 	if prevState != h.scb.State() {
 		h.info("tcp.Handler:rx-statechange", slog.Uint64("port", uint64(h.localPort)), slog.String("old", prevState.String()), slog.String("new", h.scb.State().String()), slog.String("rxflags", segIncoming.Flags.String()))
 	}
+	if segIncoming.DATALEN != 0 && h.shutdownRx && (h.scb.State() == StateFinWait1 || h.scb.State() == StateFinWait2) {
+		// soypat/lneto#50: the application is done in both directions — read side
+		// shut down (CloseRead) and our FIN sent (Close) — so inbound data has no
+		// consumer. Reply RST instead of the silent ACK-and-drop that leaves the
+		// peer waiting; the connection is torn down once the RST is sent.
+		h.scb.QueueRST(segIncoming.ACK)
+		return nil
+	}
 	if segIncoming.DATALEN != 0 && !h.shutdownRx {
 		_, err = h.bufRx.Write(payload)
 		if err != nil {
@@ -328,6 +336,10 @@ func (h *Handler) Send(b []byte) (int, error) {
 	closedSuccess := prevState == StateTimeWait && segment.Flags.HasAny(FlagACK)
 	if closedSuccess {
 		h.reset(0, 0, 0)
+	} else if segment.Flags.HasAny(FlagRST) {
+		// A sent RST aborts the connection: tear down local state now that the
+		// reset has been written to the wire (frame already in b).
+		h.Abort()
 	}
 	return datalen, nil
 }
--- a/tcp/handler.go
+++ b/tcp/handler.go
@@ -193,6 +193,7 @@ func (h *Handler) Recv(incomingPacket []byte) error {
 		// shut down (CloseRead) and our FIN sent (Close) — so inbound data has no
 		// consumer. Reply RST instead of the silent ACK-and-drop that leaves the
 		// peer waiting; the connection is torn down once the RST is sent.
+		h.info("tcp.Handler:rst-data-after-fullclose", slog.Uint64("lport", uint64(h.localPort)), slog.Uint64("rport", uint64(h.remotePort)), slog.Uint64("datalen", uint64(segIncoming.DATALEN)))
 		h.scb.QueueRST(segIncoming.ACK)
 		return nil
 	}
--- a/tcp/handler.go
+++ b/tcp/handler.go
@@ -162,7 +162,7 @@ func (h *Handler) Recv(incomingPacket []byte) error {
 		return lneto.ErrMismatch
 	}
 	payload := tfrm.Payload()
-	if len(payload) > h.bufRx.Free() {
+	if !h.shutdownRx && len(payload) > h.bufRx.Free() {
 		return lneto.ErrBufferFull
 	}
 	segIncoming := tfrm.Segment(len(payload))
--- a/tcp/handler.go
+++ b/tcp/handler.go
@@ -14,7 +14,7 @@ import (
 // related to data buffering, frame sequencing and connection state handling.
 // Does NOT implement IP related logic, so no CRC calculation/validation or pseudo header logic.
 //
-// See [Conn] for a higher level abstraction of a TCP connection, and see [ControlBlock] for the lower level bits of a TCP connection.
+// See [Conn] for a higher level abstraction of a TCP connection, and see [ControlBlock] for the low level state machine of a TCP connection.
 type Handler struct {
 	connid uint64
 	scb    ControlBlock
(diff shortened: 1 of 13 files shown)


MDr164:
The code looks pretty solid to me but my review bot (coderabbit) that I have locally found some inconsistencies. I'm not too sure yet if those are false positives so I'll have to compare the bots output with the rest of the codebase.

soypat:
Do share! This was largely authored by Claude. Want to look at it more in depth

MDr164:
I reran it through GPT5.5 to provide more 'human readable' output, to be honest I think it's fine for the scope of the issue. I pasted the output in case you want to have a look.

<details>
<summary>Verbatim Clanker Output</summary>
Main concern: Close() semantics are inconsistent

The PR comments say that after Close() “there is no reader.” That is not what the current code model says.

ControlBlock.Close() only documents that users should not send more data after close; it does not say receiving is disabled. The state model also explicitly says FIN-WAIT-1 and FIN-WAIT-2 are receive-open states via RxDataOpen(). Handler.Read() uses RxDataOpen() to decide whether EOF should be returned, so after local close and entry into FIN-WAIT-2, the API still appears to allow reads.

That also aligns with RFC 9293’s normal TCP close model: CLOSE means “I have no more data to send,” and the endpoint may continue receiving until the remote side closes. RFC 9293 also says FIN-WAIT-2 continues with established-state processing, and segment data may be delivered to receive buffers in that state.

So the PR is fine only if lneto intentionally wants Close() to mean full-duplex application close, not TCP half-close. If so, the API/docs/state helpers should be updated to reflect that. Otherwise, this PR breaks valid users that call Close() after sending a response but still expect to read peer data until FIN.

Second concern: after sending RST, the local TCB remains alive

The PR queues a RST, but nothing in ControlBlock.Send() transitions FIN-WAIT-2 to closed after that RST is sent. The send path clears pending flags and advances sequence state, but there is no FIN-WAIT-2 + RST teardown case.

At the Handler level, reset/cleanup after sending only happens for the TIME-WAIT + ACK success case; FIN-WAIT-2 + RST is not cleaned up. IsTxOver() also only treats CLOSED or TIME-WAIT with no pending segment as over, not FIN-WAIT-2 after an emitted RST.

That means the peer gets a RST, but the local connection may remain stuck in FIN-WAIT-2 unless some other cleanup path exists outside this PR. I would add an explicit post-RST teardown policy and a test for it.

Minor concern: RST generation assumes ACK is present

For the normal HTTP/browser case this is fine: established TCP data segments carry ACK, and RFC reset generation says that if the ACK bit is on, the reset should use SEQ=SEG.ACK and CTL=RST. The PR does exactly that via rstPtr = seg.ACK.

But the code does not guard against an in-window data segment without ACK. RFC 9293 says in synchronized states, if the ACK bit is off, the segment is dropped. This is probably not a real-world issue for the target bug, but it is worth either guarding or documenting.

Before merging, I would want one of these decisions made explicit:

Option A: redefine Close() as full close

Then update:
Close() docs to state that future inbound data will be reset, not read.
RxDataOpen() or Handler.Read() behavior so FIN-WAIT-2 after local Close() does not look receive-open.
Handler/TCB lifecycle so sending this RST closes or resets local state.
Tests to assert post-RST state/cleanup.

Option B: preserve TCP half-close semantics

Then keep FIN-WAIT-2 data ACK/receive behavior in ControlBlock, and move the “RST future data after app is done” policy to a higher-level Handler/Conn flag, or add a real distinction:

ShutdownWrite() / half-close: send FIN, continue reading.
Close() / full application close: send FIN, reject/reset future data.
</details>

soypat:
Aha, took me a while to understand but I see what is wrong. Will revert changes but catch the rxshutdown+close case and correctly emit RST and also terminate TCB on RST emission

tcp/handler_test.go
```diff
@@ -1155,6 +1155,146 @@ func TestHandler_RetransmitAfterMultipleLossesBothDirections(t *testing.T) {
...
+		t.Fatal("server send data:", err)
+	}
+	if n <= sizeHeaderTCP {
+		t.Fatal("server emitted no data segment")
+	}
+	return n
+}
+
+// TestFinWait2_FullClose_RST: when the application is done in BOTH directions —
+// read side shut down (CloseRead/ShutdownRead) AND our FIN sent (Close) — inbound
```
MDr164: We have ShutdownRead but do we have CloseRead?

tcp/handler.go
MDr164, line 167: I would extend this condition like so:
```go
if !h.shutdownRx && len(payload) > h.bufRx.Free() {
    return lneto.ErrBufferFull
}
```

The full-close RST path should probably not depend on the RX-buffer capacity. Or move that check elsewhere like below the full-close handling?

tcp/handler_test.go
soypat: Good catch!

soypat:
@MDr164 Addressed both of your notes and also gave documentation a revamp since I was in the neighborhood

MDr164 (approved):
lgtm

Nitpick regarding the documentation maybe: Conn.CloseRead() still says incoming data is ACKed normally and the write side is unaffected. That is true for CloseRead() alone, but after CloseRead() plus Close(), inbound data can now cause RST/teardown.

Not a big deal tbh.

soypat:
@MDr164  thanks for review!

PR #128, opened 2026-06-18
persist issue #57 fix in test

soypat:
Persists test for #57. Closes #57.

Also fixes go fix error printing in CI to make it clear what the issue is

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - name: go fix
         run: |
-          diff=$(go fix -diff ./... 2>&1)
+          diff=$(go fix -diff ./... 2>&1) || true
           if [ -n "$diff" ]; then
             echo "$diff"
             echo "::error::Files need go tool fix. Run: go tool fix ./..."
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - name: go fix
         run: |
-          diff=$(go fix -diff ./... 2>&1) || true
+          diff=$(go fix -diff ./... 2>&1)
           if [ -n "$diff" ]; then
             echo "$diff"
             echo "::error::Files need go tool fix. Run: go tool fix ./..."
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,7 +35,7 @@ jobs:
 
       - name: go fix
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
 
       - name: go fix
         run: |
-          diff=$(go fix -diff ./...)
```
MDr164, line 38: I think we could also just make that step pass regardless instead of the || true or what's the reason you intend to make this section pass either way?

tcp/handler_test.go
```diff
@@ -1154,3 +1154,99 @@ func TestHandler_RetransmitAfterMultipleLossesBothDirections(t *testing.T) {
...
+// TestRetransmit_CumulativeACK_NoSpurious reproduces soypat/lneto#57.
+// lneto streams 4 segments (TX queue=4); the first is "lost". A Linux-style
+// remote buffers the rest out of order and dup-ACKs the hole. After 3 dup ACKs
+// lneto fast-retransmits the lost segment; the remote then cumulatively ACKs
+// ALL data it received. On the buggy design (snd.NXT rewound to UNA on
+// retransmit) lneto treated that ACK as acknowledging unsent data, dropped it,
+// and emitted spurious retransmissions of already-acked segments.
+// Correct behaviour: the cumulative ACK is accepted and nothing further is sent.
+//
+// Note: an lneto Handler cannot act as the server here (it rejects out-of-order
```
MDr164, line 1167: That actually bit me while testing congestion control and I got an initial setup in which ooo packets work but it requires allocations so I'm still testing and tweaking it. I suppose for now this is fine as a test case.

.github/workflows/ci.yaml
MDr164: Nevermind, I think I misunderstood what the || true does that early next to the diff= instead of having it later in the step.

soypat: Yeah, I'm not even sure myself why we need the ||, think the main contribution is the stderr piping

soypat: will test behaviour on CI

MDr164: I think it will bail out too fast and not go into the if condition if the call inside diff= returns an error. At least that's my understanding.

MDr164 (approved):

.github/workflows/ci.yaml
soypat: you are correct, test shows it is needed

Issue #99, opened 2026-04-24
error running xcurl


jwetzell:
built using `go build ./examples/xcurl` and ran as follows.
```
~/Projects/lneto (main) » sudo ./xcurl -i ens18 -host jwetzell.com
[sudo] password for jwetzell: 
softrand 1777056631
NIC hardware address: bc:24:11:c5:c7:85 bridgeHW: bc:24:11:c5:c7:85 mtu: 1500 addr: 10.0.0.87/24
OUT 334 [Ethernet len=14; destination=ff:ff:ff:ff:ff:ff; source=us | IPv4 len=20; (Type of Service)=0x00; flags=0x4000; source=us; destination=255.255.255.255 | UDP len=8; (Source port)=68; (Destination port)=67 | DHCPv4 len=240; op=1; Flags=0x0000; (Client Address)=us; (Offered Address)=us; (Server Next Address)=255.255.255.255; (Relay Agent Address)=us; (Client Hardware Address)=us]
IN  342 [Ethernet len=14; destination=us; source=f4:1e:57:c1:6a:34 | IPv4 len=20; (Type of Service)=0x00; flags=0x0000; source=10.0.0.1; destination=10.0.0.87 | UDP len=8; (Source port)=67; (Destination port)=68 | DHCPv4 len=240; op=2; Flags=0x0000; (Client Address)=us; (Offered Address)=10.0.0.87; (Server Next Address)=10.0.0.1; (Relay Agent Address)=us; (Client Hardware Address)=us]
OUT 326 [Ethernet len=14; destination=ff:ff:ff:ff:ff:ff; source=us | IPv4 len=20; (Type of Service)=0x00; flags=0x4000; source=us; destination=255.255.255.255 | UDP len=8; (Source port)=68; (Destination port)=67 | DHCPv4 len=240; op=1; Flags=0x0000; (Client Address)=us; (Offered Address)=10.0.0.87; (Server Next Address)=10.0.0.1; (Relay Agent Address)=us; (Client Hardware Address)=us]
[...]

PR #129, opened 2026-06-18
Add xcurl port and debug flag

MDr164:
Update the docs and add new flags to xcurl to aid debugging under an OS like Linux. This is a workaround for issue #99 but no real fix.

### DIFF
--- a/examples/xcurl/main.go
+++ b/examples/xcurl/main.go
@@ -50,6 +50,8 @@ func run() (err error) {
 		flagUseHTTP       = false
 		flagHostToResolve = ""
 		flagRequestedIP   = ""
+		flagLocalTCPPort  = 0
+		flagDebugTCP      = false
 		flagDoNTP         = false
 		flagNoPcap        = false
 		flagPprof         = false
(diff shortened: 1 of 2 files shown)


examples/xcurl/main.go
```diff
@@ -351,6 +384,46 @@ func clear(buf []byte) {
...
+		tfrm.SourcePort(),
+		tfrm.DestinationPort(),
+		flags,
+		tfrm.Seq(),
+		tfrm.Ack(),
+		tfrm.WindowSize(),
+		tcpPayloadLen,
+		crcOK,
+	)
+}
```
soypat: This could be a xnet.CapturePrinter.PrintEthernet call, no?

MDr164: Yes, I quickly hacked that together to test what's going wrong and initially only wanted to keep the -port flag but figured someone else might find use in the debug stuff as well. I'll polish it up in a bit.

MDr164: Done

examples/xcurl/main.go
```diff
@@ -165,6 +191,9 @@ func run() (err error) {
 				return err
 			}
 			_, err = os.Stdout.Write(pfbuf)
+			if flagDebugTCP {
+				tcpDebug.PrintEthernet("TCP "+context, pkt)
```
soypat: I just realized we actually go through the trouble of printing out packet capture up here. tcpDebug.PrintEthernet is completely redundant with everything up here if I'm not mistaken. Why exactly did you want to print out frame information here?

MDr164: I wanted to figure out went wrong in issue #99 and just printed everything I suspected could go wrong without any particular target. As mentioned in the beginning of the PR the main thing is the README addition and the `-port` flag. The debug stuff was just left in by me IF someone needs it. I can also drop that section completely :)

soypat (approved):

PR #130, opened 2026-06-19
Retransmit active-open SYN after RTO

TuteMthCD:
## Summary

  Add SYN retransmission while a TCP connection is in `SYN-SENT`.

  Previously, lneto emitted the initial active-open SYN once and then waited for a SYN-ACK until the caller timeout expired. If that first SYN was dropped, the
  connection remained stuck in `SYN-SENT`.

  ## Change

  - Track the last SYN transmission time and SYN RTO in `tcp.Handler`.
  - Retransmit the same SYN with the same initial sequence number after the RTO expires.
  - Use exponential backoff from `1s` up to `4s`.
  - Add a regression test for SYN retransmission behavior.

  ## Validation

  ```sh
  go test ./tcp
  go test ./x/xnet

### DIFF
--- a/x/xnet/stack-blocking.go
+++ b/x/xnet/stack-blocking.go
@@ -13,8 +13,7 @@ import (
 )
 
 const (
-	maxIter              = 1000
-	controlRetryInterval = time.Second
+	maxIter = 1000
 )
 
 var (
@@ -182,8 +181,15 @@ func (s StackBlocking) DoDialTCP(conn *tcp.Conn, localPort uint16, addrp netip.A
 	if err != nil {
 		return err
 	}
+	err = s.waitDialTCP(conn, timeout)
+	if err != nil {
+		conn.Abort()
+	}
+	return err
+}
+
+func (s StackBlocking) waitDialTCP(conn *tcp.Conn, timeout time.Duration) (err error) {
 	deadline := time.Now().Add(timeout)
-	controlRetryAt := time.Now().Add(controlRetryInterval)
 	var backoffs uint
 	for range maxIter {
 		state := conn.State()
@@ -191,16 +197,10 @@ func (s StackBlocking) DoDialTCP(conn *tcp.Conn, localPort uint16, addrp netip.A
 			return nil
 		} else if state == tcp.StateSynSent || state == tcp.StateSynRcvd || conn.AwaitingSynSend() {
 			if err = s.checkDeadline(deadline); err != nil {
-				conn.Abort()
 				return err
-			}
-			if conn.IsAwaitingControl() && time.Since(controlRetryAt) >= 0 {
-				conn.RequeueControl()
-				controlRetryAt = time.Now().Add(controlRetryInterval)
 			}
 		} else {
 			// Unexpected state, abort and terminate connection.
-			conn.Abort()
 			return errTCPFailedToConnect
 		}
 		s.backoff(backoffs)
--- a/x/xnet/stack-blocking.go
+++ b/x/xnet/stack-blocking.go
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	maxIter = 1000
+	maxIter              = 1000
+	controlRetryInterval = time.Second
 )
 
 var (
@@ -182,6 +183,7 @@ func (s StackBlocking) DoDialTCP(conn *tcp.Conn, localPort uint16, addrp netip.A
 		return err
 	}
 	deadline := time.Now().Add(timeout)
+	controlRetryAt := time.Now().Add(controlRetryInterval)
 	var backoffs uint
 	for range maxIter {
 		state := conn.State()
@@ -191,6 +193,10 @@ func (s StackBlocking) DoDialTCP(conn *tcp.Conn, localPort uint16, addrp netip.A
 			if err = s.checkDeadline(deadline); err != nil {
 				conn.Abort()
 				return err
+			}
+			if conn.IsAwaitingControl() && time.Since(controlRetryAt) >= 0 {
+				conn.RequeueControl()
+				controlRetryAt = time.Now().Add(controlRetryInterval)
 			}
 		} else {
 			// Unexpected state, abort and terminate connection.
--- a/x/xnet/stack-go.go
+++ b/x/xnet/stack-go.go
@@ -175,7 +175,6 @@ func (s StackGo) SocketNetip(ctx context.Context, network string, family, sotype
 			if err != nil {
 				return nil, err
 			}
-			controlRetryAt := time.Now().Add(controlRetryInterval)
 			var backoffs uint
 			for {
 				s.blk.backoff(backoffs)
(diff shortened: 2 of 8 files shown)


MDr164:
Generally speaking it's a good addition but it introduces the time package into TCP. I'm also working on a similar item and right now trying to keep the time additions optional. Other than that it looks fine.

soypat:
It seems this belongs at a different level of abstraction in lneto's case. See https://github.com/soypat/lneto/blob/6f5acd1948e20299790bcb1dead9fc4e6fb284f7/x/xnet/stack-retrying.go#L93-L110 which according to my understanding technically already does SYN RTO

TuteMthCD:
> It seems this belongs at a different level of abstraction in lneto's case. See
> 
> https://github.com/soypat/lneto/blob/6f5acd1948e20299790bcb1dead9fc4e6fb284f7/x/xnet/stack-retrying.go#L93-L110
> which according to my understanding technically already does SYN RTO

(AI-assisted clanker English)

My understanding is that SYN retransmission should be handled by the TCP layer rather than by Dial. StackRetrying.DoDialTCP retries the entire connection attempt, but it does not retransmit packets within the same TCP attempt. For that reason, I implemented the change at the TCP level to retransmit the SYN when no response is received.

I understand that StackRetrying.DoDialTCP can address some cases by retrying the whole connection from a higher layer. However, I see these as different mechanisms: a Dial retry creates a new connection attempt, whereas a SYN retransmission aims to complete the same TCP handshake correctly.

For that reason, it felt more natural to handle this case within TCP. From my perspective, SYN retransmission was not previously handled within the TCP implementation, which is why I added that logic there.

Regarding the use of time, the goal was to add SYN retransmission at the TCP layer. If you'd prefer to avoid that dependency, what mechanism would you recommend for implementing this kind of retransmission? That way I can adapt the solution to align with the project's preferred approach.

soypat:
Yes, it seems more natural to handle it in tcp package, after all that is what every other implementation does out there. That said, It is desirable to keep timing logic out of tcp package if possible, and it seems entirely possible in this case. 

To address your concern of keeping the same connection on retry we can add a method on tcp.Conn that retries sending a control segment out if any are pending. An example of how this could look:

```go
// If stuck on synsend or synwait will retry sending out a control segment
// to advance connection state in case of packet loss. conn is tcp.Conn
if conn.IsAwaitingControl() && time.Since(deadline)>=0 {
    conn.RequeueControl() // Or a better name if you can think of one.
    continue
}
```
This change would belong in StackBlocking.DoDialTCP

TuteMthCD:
• Updated the approach based on the review feedback.

  The TCP package no longer owns retransmission timing. Instead, it only exposes control-segment state and requeue APIs:

  - `IsAwaitingControl()`
  - `RequeueControl()`

  The retry timing now lives in the dialing loops:

  - `StackBlocking.DoDialTCP`
  - `StackGo.SocketNetip`

  When a TCP dial remains in `SYN-SENT` / `SYN-RCVD`, the dial loop periodically calls `RequeueControl()`, allowing the next send to retransmit the pending
  control segment without embedding timer logic inside the TCP package.

  I also updated the regression test to cover explicit control requeue behavior.

x/xnet/stack-blocking.go
```diff
@@ -192,6 +194,10 @@ func (s StackBlocking) DoDialTCP(conn *tcp.Conn, localPort uint16, addrp netip.A
 				conn.Abort()
 				return err
 			}
+			if conn.IsAwaitingControl() && time.Since(controlRetryAt) >= 0 {
+				conn.RequeueControl()
+				controlRetryAt = time.Now().Add(controlRetryInterval)
+			}
```
soypat: silly me, I realize maybe we could put this in StackRetries so that we use the user parameters of number of retries and retry interval as provided by user. One thing to note is if the connection is aborted then you'd need to open the connection again instead of queuing control retry.

x/xnet/tcppool.go
```diff
@@ -42,8 +42,7 @@ type TCPPoolConfig struct {
 	ConnLogger *slog.Logger
 
 	// NanoTime returns the current monotonic time in nanoseconds.
-	// Used for pool timeout tracking and passed to each [tcp.Conn] for
-	// retransmission timing (RFC 6298). If nil, defaults to time.Now().UnixNano().
```
soypat, line 46: was this on purpose?

x/xnet/stack-blocking.go
```diff
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	maxIter = 1000
+	maxIter              = 1000
+	controlRetryInterval = time.Second
```
soypat: if possible add a RFC reference or paper for the controlRetryInterval

x/xnet/tcppool.go
TuteMthCD: Yes and no. I updated the comment in one implementation but forgot to update it in the other one. :/ 
😅

x/xnet/stack-go.go
```diff
@@ -171,9 +177,16 @@ func (s StackGo) SocketNetip(ctx context.Context, network string, family, sotype
 			if err != nil {
 				return nil, err
 			}
-			err = s.blk.async.DialTCP(&conn, laddr.Port(), raddr)
-			if err != nil {
-				return nil, err
+			if s.tcpDialRetries > 0 && s.tcpDialTimeout > 0 {
+				err = (StackRetrying{block: s.blk}).DoDialTCP(&conn, laddr.Port(), raddr, s.tcpDialTimeout, s.tcpDialRetries)
```
soypat: StackRetrying is a method on StackBlocking, use that instead of creating manually. 

Also if when creating a StackRetrying the values have sane defaults (as configured above) you can avoid having to check here if tcpDialRetries and tcpDialTimeout have non-zero values and just call DoDialTCP directly and omit this big if statement.

soypat (commented):
looks much better! A few notes that may lead to improved code

x/xnet/stack-go.go
```diff
@@ -33,15 +35,19 @@ func (s *StackAsync) StackGo(stackProtoBackoff lneto.BackoffStrategy, cfg StackG
 
 func (s StackBlocking) StackGo(cfg StackGoConfig) StackGo {
 	sg := StackGo{
-		blk:   s,
-		plcfg: cfg.ListenerPoolConfig,
+		blk:            s,
+		plcfg:          cfg.ListenerPoolConfig,
+		tcpDialTimeout: cfg.TCPDialTimeout,
+		tcpDialRetries: cfg.TCPDialRetries,
```
soypat: Had left a comment here but it dissapeared. Basically set sane default values:

if cfg.dialretries<=0: tcpdialtimeout=1
if tcpdialtimeout<=0: tcpdialtimeout=2s

soypat:
Failing test and failing formatting. You've revealed an issue in CI, to show the needed diff in formatting run:
```
gofmt -d .
```
https://github.com/soypat/lneto/pull/133

soypat (approved):

PR #131, opened 2026-06-19
Add atomic id guard to prevent TOCTOU

### LINKED ISSUE #82: Parallel tcp.Conn.Close | tcp.Conn.Write calls cause ongoing write data to be dropped.
Situation:
Two ongoing goroutines sharing the same tcp.Conn
- G1 Calls Write with more data than Handler Tx buffer has free.
- G2 calls Close immediately after.

This causes G1 to return after first portion of buffer written is sent over the wire since the Handler buffer is emptied and G1 may have been unable to fill it before

Probably not worth fixing with current abstractions- instead developing a more intricate io_uring implementation to better support complex parallel IO operations.

MDr164:
This resolves #82 as suggested by @soypat to use a write lock via a atomic uint64.

### DIFF
--- a/tcp/conn.go
+++ b/tcp/conn.go
@@ -8,6 +8,7 @@ import (
 	"net/netip"
 	"os"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/soypat/lneto"
@@ -29,6 +30,14 @@ type Conn struct {
 	h          Handler
 	remoteAddr []byte
 
+	// writeLock holds the connection ID ([Handler.connid]) of the goroutine
+	// that currently owns the write side via [Conn.Write] or [Conn.Flush].
+	// A zero value means no write is in progress. It serializes concurrent
+	// writers and lets [Conn.Close] wait for an in-flight write to finish
+	// queueing its data instead of dropping it (see issue #82). It is an atomic
+	// so Close can poll it without contending on mu while a writer drains.
+	writeLock atomic.Uint64
+
 	_backoff lneto.BackoffStrategy
 	rdead    time.Time
 	wdead    time.Time
@@ -37,6 +46,12 @@ type Conn struct {
 
 	ipID uint16
 }
+
+// maxCloseDrainBackoffs bounds how many backoff iterations [Conn.Close] waits
+// for an in-progress write to drain before closing regardless. It is a safety
+// net against a writer that can never make progress (e.g. a dead peer with no
+// write deadline); a writer with a deadline releases the lock far sooner.
+const maxCloseDrainBackoffs = 1 << 16
 
 // reset must be called while holding [Conn.mu].
 func (conn *Conn) reset(h Handler) {
@@ -49,6 +64,7 @@ func (conn *Conn) reset(h Handler) {
 	conn.wdead = time.Time{}
 	conn.abortErr = nil
 	conn.ipID = 0
+	conn.writeLock.Store(0)
 }
 
 type ConnConfig struct {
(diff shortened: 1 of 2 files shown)


tcp/conn.go
```diff
@@ -337,14 +367,42 @@ func (conn *Conn) Read(b []byte) (int, error) {
...
-		return 0, err
+// beginWrite validates the connection is open and acquires the write lock for
+// the current connection, returning its connection ID. It serializes concurrent
+// writers: while another goroutine owns the write side it blocks with backoff
+// until that writer releases or the connection closes or the write deadline
+// elapses. The returned connID must be released with [Conn.endWrite].
+//
+// The lock is acquired while holding mu so that a concurrent [Conn.Close]
+// observing writeLock cannot race past an in-progress write.
+func (conn *Conn) beginWrite() (connID uint64, err error) {
```
soypat: Call this acquireWriteLock

tcp/conn.go
```diff
@@ -337,14 +367,42 @@ func (conn *Conn) Read(b []byte) (int, error) {
...
+			return 0, errDeadlineExceeded
+		}
+		conn.backoff(backoffs)
+		backoffs++
 	}
-	return conn.h.connid, nil
+}
+
+// endWrite releases a write lock previously acquired by [Conn.beginWrite].
+func (conn *Conn) endWrite(connID uint64) {
```
soypat: and call this releaseWriteLock

tcp/conn.go
```diff
@@ -337,14 +367,42 @@ func (conn *Conn) Read(b []byte) (int, error) {
...
+// the current connection, returning its connection ID. It serializes concurrent
+// writers: while another goroutine owns the write side it blocks with backoff
+// until that writer releases or the connection closes or the write deadline
+// elapses. The returned connID must be released with [Conn.endWrite].
+//
+// The lock is acquired while holding mu so that a concurrent [Conn.Close]
+// observing writeLock cannot race past an in-progress write.
+func (conn *Conn) beginWrite() (connID uint64, err error) {
+	var backoffs uint
+	for {
```
soypat, line 366: I think this function would look much cleaner if you start by reading the current connID and using checkPipe instead of holding the lock manually.

```go
// simple load should be OK for our purposes? See https://research.swtch.com/gomm
connID := conn.h.connid 
var backoffs uint
for {
   if err := conn.checkPipe(connID, &conn.wdead); err != nil {
      return 0, err
    } else if conn.writeLock.CompareAndSwap(0, connID) {
       return connID, nil
    } 
    conn.backoff(backoffs)
    backoffs++
}
```

tcp/conn.go
```diff
@@ -201,6 +217,18 @@ func (conn *Conn) CloseRead() error {
...
+	conn.mu.Unlock()
+	// Wait for an in-progress Write/Flush on this connection to finish queueing
+	// its data so that closing does not silently drop it (issue #82). The wait is
+	// bounded so a writer that can never make progress does not block Close forever.
+	for backoffs := uint(0); connid != 0 && backoffs < maxCloseDrainBackoffs; backoffs++ {
+		if conn.writeLock.Load() != connid {
+			break // No writer owns this connection's write side.
+		}
+		conn.backoff(backoffs)
+	}
```
soypat: This is kind of odd and I see why you've added maxCloseDrainBackoffs but I feel it is off. The connID semantics during the close/abort/reset phase are not very well defined so this feels like it could break in the future.  Would waiting to acquire write lock and hold it throughout Close not solve the issue?

soypat (commented):
potential simplifications to improve code?

tcp/conn.go
MDr164: True, the current code is a bit overcomplicated. I slimmed it down without touching the new test and it still passes.

soypat:
looks absolutely wunderbar. thank you @MDr164 :)

PR #132, opened 2026-06-20
draft DHCP allocator abstraction

soypat:
Addressing https://github.com/tinygo-org/espradio/pull/51#discussion_r3441273645

```go
type Allocator interface {
	// Offer tentatively reserves a binding for the requesting client and
	// returns it. It is called when the server receives a DISCOVER/SOLICIT.
	Offer(Request) (Binding, error)
	// Commit binds a previously offered reservation, finalizing the lease, and
	// returns the committed binding. It is called when the server receives a
	// REQUEST.
	Commit(Request) (Binding, error)
	// Release frees the binding held for the given client identity. It is
	// called when the server receives a RELEASE. Releasing an unknown client is
	// not an error.
	Release(clientID []byte) error
	// Decline marks the addresses in the request as unusable, typically because
	// the client detected an address conflict. It is called when the server
	// receives a DECLINE.
	Decline(Request) error

	// AppendOptions lets the allocator customize the option bytes the server is
	// about to send. dst already contains the options the server derived from
	// its own configuration (server identifier, router, subnet mask, DNS, lease
	// times, ...) for the given client and binding. The implementation may
	// append further options, or rewrite the existing ones, and must return the
	// resulting slice. The returned slice must not exceed dst's capacity.
	AppendOptions(dst []byte, clientID []byte, b Binding) ([]byte, error)
}
```

### DIFF
--- a/dhcp/dhcpv4/allocator.go
+++ b/dhcp/dhcpv4/allocator.go
@@ -1,0 +1,227 @@
+package dhcpv4
+
+import (
+	"encoding/binary"
+	"errors"
+	"net/netip"
+	"time"
+
+	"github.com/soypat/lneto/dhcp"
+	"github.com/soypat/lneto/ipv4"
+)
+
+// MapAllocator is the built-in [dhcp.Allocator] used by [Server] when no
+// allocator is supplied. It keeps the lease database in a map and allocates
+// addresses sequentially from the configured subnet, preferring a client's
+// requested address when it is free and in-subnet.
+//
+// When configured with a clock (see [AllocatorConfig.Now]) it reclaims expired
+// leases back into the pool; with no clock leases never expire, matching the
+// behaviour of a server that does not track time.
+//
+// The map makes MapAllocator unsuitable for the most memory-constrained
+// targets; such deployments should provide their own [dhcp.Allocator]. This
+// mirrors the existing allowance for the DHCP server to use a map on capable
+// hardware.
+type MapAllocator struct {
+	subnet       ipv4.Prefix
+	serverAddr   [4]byte
+	nextAddr     [4]byte
+	leaseSeconds uint32
+	now          func() time.Time
+	leases       map[[36]byte]mapLease
+	declined     map[[4]byte]struct{}
+}
+
+type mapLease struct {
+	addr      [4]byte
+	expiry    time.Time // zero means no expiry (no clock configured).
+	committed bool
+}
+
+// AllocatorConfig configures a [MapAllocator].
+type AllocatorConfig struct {
+	// ServerAddr is the server's own address, excluded from the pool.
+	ServerAddr [4]byte
+	// Subnet is the allocation prefix.
+	Subnet ipv4.Prefix
+	// LeaseSeconds is the lease duration handed to clients. Zero defaults to 3600.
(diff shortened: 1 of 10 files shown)


MDr164:
Quickly glanced over it and already found a couple questionable items in this PR. Want me to review or rather wait until you moved it out of draft?

soypat:
As you wish, this PR is pure clankerspeak. I'll eventually take a more in depth look, just wanted to get it out there so I could see diff comfortably

dhcp/dhcpv4/allocator.go
```diff
@@ -0,0 +1,227 @@
...
+	if a.nextAddr == a.serverAddr {
+		a.nextAddr = a.subnet.Next(a.nextAddr)
+	}
+	// Reject the broadcast address (all host bits set).
+	hostBits := uint(32 - a.subnet.Bits())
+	hostMask := ^uint32(0) >> (32 - hostBits)
+	if binary.BigEndian.Uint32(addr[:])&hostMask == hostMask {
+		return [4]byte{}, false
+	}
+	return addr, true
```
MDr164, line 205: Am I missing anything or should there not be a `!a.isAssigned` check in this section as well?
So the sequential fallback here returns nextAddr without applying the same check as above.

For example, if one client requests .10, the sequential allocator can later reach .10 and hand it to another client.

There are also network/broadcast edge cases: a client can explicitly request either address, and after the sequential path rejects the broadcast address, nextAddr has already wrapped and the next call can return the network address.

Could this use one shared usable(addr) check and scan candidates until it either finds a usable address or returns to its starting point?

dhcp/dhcpv4/allocator.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func keyOf(clientID []byte) ([36]byte, bool) {
+	var key [36]byte
+	if len(clientID) > len(key) {
+		return key, false
+	}
+	copy(key[:], clientID)
+	return key, true
+}
```
MDr164, line 227: The key does not include the client ID length. Since the remaining bytes are zero-filled, IDs such as:

```go
[]byte{1}
[]byte{1, 0}
```

produce the same map key I think.

Could the key include the length, for example:

```go
type clientKey struct {
    id  [36]byte
    len uint8
}
```

dhcp/dhcpv4/allocator.go
```diff
@@ -0,0 +1,227 @@
...
+	leaseSeconds uint32
+	now          func() time.Time
+	leases       map[[36]byte]mapLease
+	declined     map[[4]byte]struct{}
+}
+
+type mapLease struct {
+	addr      [4]byte
+	expiry    time.Time // zero means no expiry (no clock configured).
+	committed bool
```
MDr164, line 39: committed is set during Commit, but I cannot find anywhere that reads it, did I miss it?

Was this intended to distinguish tentative reservations from committed leases, perhaps with different expiration behavior? Otherwise it can probably be removed.

MDr164 (commented):
There are a couple of other things probably but I ended digging deeper as it's still a draft.

soypat:
@MDr164 Thank you for your review. Will revisit this when I find the strength. Was aware the code was not quite solid when I let the clanker plough through

PR #133, opened 2026-06-22
try to fix fmt msg

soypat:


soypat:
Before: https://github.com/soypat/lneto/actions/runs/27983466870/job/82819238055?pr=130
After: https://github.com/soypat/lneto/actions/runs/27984370544/job/82822166248

PR #134, opened 2026-06-23
Add tinygo and fix CI failure

MDr164:
We previously did not run any tinygo tests and one test fails on main right now which this change should fix.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,6 +64,11 @@ jobs:
         with:
           go-version: "1.26"
 
+      - uses: acifani/setup-tinygo@dd8a7075d951a7595b2ef2123ed0ab1af0c13e56 # v3.0.0
+        with:
+          tinygo-version: "0.41.1"
+          install-binaryen: "false"
+
       - name: Test
         run: |
           mkdir -p profiles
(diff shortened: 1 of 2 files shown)


soypat:
You pointed out an important source of flakiness but I was not convinced of the validity of the failing test due to the stochastic nature of the failing test. Limiting maxIter is a good idea but I wanted to figure out a way to test that took real operating conditions into account. Hitting maxIter is not very realistic in practice. So I opted for approaching the problem with a strict scheduling in test using channels which should completely eliminate flakiness and also speed up tests considerably.

https://github.com/soypat/lneto/pull/135

soypat:
Also lets discuss adding tinygo to CI. Adding dependencies to CI is a touchy subject for me :sweat_smile:

MDr164:
Fair, adding tinygo was a might-as-well kinda decision here. As lneto somewhat supports it having it in the test container (no other CI job has it, just the test one) it allows executing tests that required tinygo present. Up to you, how our "support guarantees" for it are. I consume lneto via tamago personally but some coworkers had more interest in the tinygo side as they used that one more.

MDr164:
Closed in favor of #135

PR #135, opened 2026-06-23
Incorporate strict test timing for TestStackGoTCPDialRetriesPendingControl

soypat:
Test now runs with strict timing guarantees eliminating all flakiness. This also adds logical time to StackBlocking instead of relying on time package. This results in huge test speedup as well. 
```
go test  -run TestStackGoTCPDialRetriesPendingControl  -count=50000 ./x/xnet 
ok      github.com/soypat/lneto/x/xnet  2.737s
tinygo test  -run TestStackGoTCPDialRetriesPendingControl -count=50000  ./x/xnet
ok      github.com/soypat/lneto/x/xnet  29.900s
```

PR #136, opened 2026-06-24
fix: two fixes for dhcp/ap

deadprogram:
This PR contains two fixes for DHCP and UDP discovered in espradio when using as an access point:

- fix: accept 255.255.255.255 dst in demux4 
- fix(dhcp): use chaddr as lookup key and patch Ethernet dst on Offer/Ack

It also includes test coverage for these changes.

### DIFF
--- a/dhcp/dhcpv4/server_ap_test.go
+++ b/dhcp/dhcpv4/server_ap_test.go
@@ -1,0 +1,252 @@
+package dhcpv4
+
+// Tests covering the AP-mode DHCP server fixes:
+//   1. Client lookup by chaddr when no ClientID option is present.
+//   2. Ciaddr fallback lookup when chaddr lookup finds nothing.
+//   3. Ethernet dst is patched to chaddr on Offer/Ack when Encapsulate is
+//      called with offsetToIP >= 14.
+
+import (
+	"testing"
+
+	"github.com/soypat/lneto/ipv4"
+)
+
+// doraNoClientID runs a complete DORA exchange using raw DHCP frames
+// (no ClientID option) with the given chaddr. The client uses only chaddr
+// as its identity, matching normal AP-mode behaviour where mobile clients
+// don't send OptClientIdentifier.
+func doraNoClientID(t *testing.T, sv *Server, chaddr [6]byte, xid uint32) (assignedIP [4]byte) {
+	t.Helper()
+	var cl Client
+	err := cl.BeginRequest(xid, RequestConfig{
+		ClientHardwareAddr: chaddr,
+		// Intentionally no ClientID – forces server to key on chaddr.
+	})
+	if err != nil {
+		t.Fatalf("BeginRequest: %v", err)
+	}
+
+	var buf [1024]byte
+
+	// DISCOVER
+	n, err := cl.Encapsulate(buf[:], -1, 0)
+	if err != nil || n == 0 {
+		t.Fatalf("discover encapsulate: n=%d err=%v", n, err)
+	}
+	if err = sv.Demux(buf[:n], 0); err != nil {
+		t.Fatalf("discover demux: %v", err)
+	}
+
+	// OFFER
+	n, err = sv.Encapsulate(buf[:], -1, 0)
+	if err != nil || n == 0 {
+		t.Fatalf("offer encapsulate: n=%d err=%v", n, err)
+	}
+	frm, _ := NewFrame(buf[:n])
+	assignedIP = *frm.YIAddr()
+	if err = cl.Demux(buf[:n], 0); err != nil {
+		t.Fatalf("offer demux: %v", err)
+	}
+
+	// REQUEST
+	n, err = cl.Encapsulate(buf[:], -1, 0)
+	if err != nil || n == 0 {
+		t.Fatalf("request encapsulate: n=%d err=%v", n, err)
+	}
+	if err = sv.Demux(buf[:n], 0); err != nil {
+		t.Fatalf("request demux: %v", err)
+	}
+
+	// ACK
+	n, err = sv.Encapsulate(buf[:], -1, 0)
+	if err != nil || n == 0 {
+		t.Fatalf("ack encapsulate: n=%d err=%v", n, err)
+	}
+	if err = cl.Demux(buf[:n], 0); err != nil {
+		t.Fatalf("ack demux: %v", err)
+	}
+
+	if cl.State() != StateBound {
+		t.Errorf("want StateBound, got %s", cl.State())
+	}
+	return assignedIP
+}
+
+// TestServerChaddrLookup_NoClientID verifies the server can complete a full DORA
+// cycle when the client sends no OptClientIdentifier option.  Before the fix the
+// MsgRequest Demux call would fail with "request for non existing client" because
+// the server tried to look up the entry by ciaddr (0.0.0.0) instead of chaddr.
+func TestServerChaddrLookup_NoClientID(t *testing.T) {
+	svAddr := [4]byte{192, 168, 4, 1}
+	var sv Server
+	sv.Configure(ServerConfig{
+		ServerAddr: svAddr,
+		Gateway:    svAddr,
+		Subnet:     ipv4.PrefixFrom(svAddr, 24),
+	})
+
+	chaddr := [6]byte{0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff}
+	ip := doraNoClientID(t, &sv, chaddr, 0x12345678)
+	if ip[0] != 192 || ip[1] != 168 || ip[2] != 4 {
+		t.Errorf("unexpected subnet in assigned IP %v", ip)
+	}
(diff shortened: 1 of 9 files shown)


deadprogram:
The test failure here is due to https://github.com/soypat/lneto/pull/137

deadprogram:
This PR now includes the commit from #137 so it will need to be rebased after that fix has been merged.

deadprogram:
Updated to use #135 so no further rebase needed. ready for review!

internet/stack-udpport.go
```diff
@@ -45,7 +45,7 @@ func (sudp *StackUDPPort) Demux(carrierData []byte, frameOffset int) error {
 	if dst != sudp.h.lport {
 		return lneto.ErrPacketDrop // Not meant for us.
 	}
-	if len(sudp.raddr) > 0 && !internal.IsMulticastIPAddr(sudp.raddr) {
```
soypat, line 48: This is curious. What does this change solve? If the zero address is invalid it should be rejected at the IP stack level

internet/stack-ip4.go
```diff
@@ -101,7 +101,8 @@ func (si4 *stackip4) demux4(carrierData []byte, offset int) error {
 	}
 	dst := ifrm.DestinationAddr()
 	if si4.ip4 != ([4]byte{}) && *dst != si4.ip4 {
-		if !si4.acceptMulticast || !internal.IsMulticastIPAddr(dst[:]) {
+		isBroadcast := *dst == ([4]byte{255, 255, 255, 255})
```
soypat: Is it possible to rebase onto https://github.com/soypat/lneto/pull/139 ?

internet/ap_test.go
```diff
@@ -0,0 +1,285 @@
+package internet
```
soypat: Stack4/6 are mostly tested at the xnet level. These tests don't really add too much coverage with the exception of the broadcast test.

soypat: consider dropping this file entirely

dhcp/dhcpv4/server_ap_test.go
```diff
@@ -0,0 +1,567 @@
...
+	frm.SetMagicCookie(MagicCookie)
+
+	opts := frm.OptionsPayload()
+	n, _ := EncodeOption(opts, OptMessageType, byte(msgType))
+	copy(opts[n:], extraOpts)
+	n += len(extraOpts)
+	opts[n] = byte(OptEnd)
+	n++
+	return append([]byte(nil), buf[:OptionsOffset+n]...)
+}
```
soypat: Ditch UDP/IP/Ethernet level tests in dhcp package. If you want to test stack integration at the stack level consider writing tests in xnet package. Notice how we exclusively test DHCP logic in our tests in this package. Follow the style in the package.

dhcp/dhcpv4/server_ap_test.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+	if err = cl.Demux(buf[:n], 0); err != nil {
+		t.Fatalf("ack demux: %v", err)
+	}
+
+	if cl.State() != StateBound {
+		t.Errorf("want StateBound, got %s", cl.State())
+	}
+	return assignedIP
+}
```
soypat, line 74: Use a Client to simulate client interactions please.

dhcp/dhcpv4/server_ap_test.go
```diff
@@ -0,0 +1,567 @@
...
+		Gateway:    svAddr,
+		Subnet:     ipv4.PrefixFrom(svAddr, 24),
+	})
+
+	chaddr := [6]byte{0xaa, 0xbb, 0xcc, 0xdd, 0xee, 0xff}
+	ip := doraNoClientID(t, &sv, chaddr, 0x12345678)
+	if ip[0] != 192 || ip[1] != 168 || ip[2] != 4 {
+		t.Errorf("unexpected subnet in assigned IP %v", ip)
+	}
+}
```
soypat, line 94: See server_test.go for inspiration on how we test

dhcp/dhcpv4/server_ap_test.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+	udpCksum := ^uint16(udpSum)
+	if udpCksum == 0 {
+		udpCksum = 0xFFFF
+	}
+	binary.BigEndian.PutUint16(udp[6:8], udpCksum)
+
+	copy(buf[ethHdr+ipHdr+udpHdr:], dhcpPayload)
+	return total
+}
```
soypat: same as above, only dhcp in dhcp package.

dhcp/dhcpv4/server_ap_test.go
```diff
@@ -0,0 +1,567 @@
...
+	sv.Configure(ServerConfig{
+		ServerAddr: svAddr,
+		Gateway:    svAddr,
+		Subnet:     ipv4.PrefixFrom(svAddr, 24),
+	})
+
+	var cl Client
+	cl.BeginRequest(0xabcd1234, RequestConfig{
+		ClientHardwareAddr: clChaddr,
+	})
```
soypat: this is what we want to see in tests that have server/client interaction

dhcp/dhcpv4/server_ap_test.go
```diff
@@ -0,0 +1,567 @@
...
+		t.Fatalf("client not bound after DORA, state=%s", cl.State())
+	}
+	if len(sv.hosts) == 0 {
+		t.Fatal("server should have 1 client after DORA")
+	}
+
+	// Phase 2: send a RELEASE with no ClientID and ciaddr=assignedIP.
+	// The chaddr lookup will fail (entry is keyed by "persistent-device-id"),
+	// triggering the getClientByIP fallback.
+	release := makeDHCPFrame(t, MsgRelease, xid, assignedIP, chaddr, nil)
```
soypat: You've added a LOT of tests for such a simple change. consider matching tests to actual functionality added please. keep tests concise and to the point. I suggest you tell your agent to write a test for every bug/missing feature. ONE test, for each. and then after you have a failing test for each, go fixing the issues/adding features one by one.

soypat (changes requested):
Keep tests concise. Last comment has a tip I found extremely useful for writing useful, well-meaning tests with agents. The actual code changes you've provided seem valid and actual issue-fixing.

internet/stack-udpport.go
soypat: ahh if it is zeroed it skips IP address validation. But this seems like a hacky patch- the DHCP server should be registered with a zero-lengthed raddr and that would have same effect.

```go
udpStackPort.SetStackNode(node, nil, dhcpv4.DefaultServerPort)
```

deadprogram: I think it is no longer needed, thanks. Removed.

internet/stack-ip4.go
deadprogram: Done!

internet/ap_test.go
deadprogram: Dropped. :wastebasket:

dhcp/dhcpv4/server_ap_test.go
deadprogram: Removed.

dhcp/dhcpv4/server_ap_test.go
deadprogram: Done.

dhcp/dhcpv4/server_ap_test.go
deadprogram: Removed.

dhcp/dhcpv4/server_ap_test.go
deadprogram: Good point. Removed.

deadprogram:
All updated and feedback should be addressed. PTAL.

deadprogram:
Updated again to increase test coverage.

dhcp/dhcpv4/server_ap_test.go
```diff
@@ -0,0 +1,252 @@
...
+	var got [6]byte
+	copy(got[:], carrier[0:6])
+	if got != clChaddr {
+		t.Errorf("Ethernet dst: got %v, want %v", got, clChaddr)
+	}
+}
+
+// doraWithClient runs a complete DORA exchange using the given pre-configured
+// Client and returns the assigned IP.
+func doraWithClient(t *testing.T, sv *Server, cl *Client) (assignedIP [4]byte) {
```
soypat, line 211: Reuse this function in doraNoClientID (one of first functions)

soypat:
Heads up, resolved conflicts!

deadprogram:
Tests still passed. Can we squash and merge?

soypat (approved):
one of my comments was not addressed but no biggie, no sweat. looks like a swell improvement so in it goes!

soypat:
Thank you for the PR!

PR #137, opened 2026-06-24
fix(xnet): remove maxIter cap from waitDialTCP

deadprogram:
The function used `for range maxIter` (1000 iterations) as an exit guard. With a Gosched backoff, 1000 iterations complete in ~1ms, well before the configured dial timeout. When the loop exhausted, the connection was in AwaitingSynSend (SYN not yet emitted) so IsAwaitingControl() returned false causing the conn to be aborted instead of requeued. This causes the retry to start fresh and the test's EgressEthernet driver to never observe a SYN.

Replace the bounded loop with an infinite loop whose only exit paths are the deadline check, StateEstablished, or an unexpected state. The timeout is now the sole time limit, matching caller intent.

Fixes TestStackGoTCPDialRetriesPendingControl which failed deterministically without the race detector.


soypat:
Should be solved in https://github.com/soypat/lneto/pull/135! Please check

deadprogram:
Appears to be addressed by #135 so now closing. Thanks for the fix @soypat

PR #139, opened 2026-06-24
feature: ipv4 broadcast support and bugfix for reset connection reset of IPv4+IPv6 stacks

soypat:
Basically refactored a one line change in https://github.com/soypat/lneto/pull/136 into a more concious option on the networking stack.


deadprogram:
Tested as part of #136 and worked great!

soypat:
@deadprogram Thank you Ron!

Issue #140, opened 2026-06-24
Eliminate all time.Sleep/time.Now calls in tests


soypat:
The introduction of [stackTestScheduler](https://github.com/soypat/lneto/blob/a42f0b404c8934a1aa64de9a62e06fce44843d00/x/xnet/xnet_test.go#L42) abstraction provides a proper way to test networking stacks that depend on sleeping and time passing for logic. This is a source of flakiness and undesirable as a whole.

@MDr164 notes there are issues in using channels as it is currently presented, but I'm sure that nothing too serious that would impede the widespread adoption of this methodology or a similar one in lneto.

MDr164:
Yes, I'm generally in favor of using channels, I think the only issues I saw were in the test files where they prematurely accepted a condition as passing even though a delayed error might be inflight. I'll gonna give it a deeper look this evening, maybe I remembered it wrong. We could also utilize the new [synctest](https://go.dev/blog/synctest) stdlib pkg for those cases I think.

PR #142, opened 2026-06-25
EXPERIMENT: bump go to 1.25 and use synctest

MDr164:
In 1.24 synctest was added as an experiment and it got stabilized further in 1.25. This also migrates a test to it for initial evaluation.

### DIFF
--- a/x/xnet/xnet_dialretry_synctest_test.go
+++ b/x/xnet/xnet_dialretry_synctest_test.go
@@ -1,0 +1,101 @@
+//go:build !tinygo
+
+// synctest is unavailable under TinyGo, so this deterministic dial-retry test is
+// excluded from TinyGo builds. The retry/timeout logic it exercises is platform
+// independent and fully covered when run with the standard Go toolchain.
+package xnet
+
+import (
+	"context"
+	"errors"
+	"net/netip"
+	"syscall"
+	"testing"
+	"testing/synctest"
+	"time"
+
+	"github.com/soypat/lneto"
+	"github.com/soypat/lneto/ethernet"
+	"github.com/soypat/lneto/tcp"
+)
+
+// TestStackGoTCPDialRetriesPendingControl verifies that an active TCP dial fails
+// after exhausting its retries when the peer never responds to the SYN.
+//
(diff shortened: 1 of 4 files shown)


soypat:
I had started work on a better scheduler implementation. will create a PR now

x/xnet/xnet_dialretry_synctest_test.go
```diff
@@ -0,0 +1,91 @@
...
+			t.Fatal(err)
+		}
+		frm, ok := getTCPFrame(buf[:n])
+		if !ok {
+			t.Fatal("expected an outbound TCP SYN from dial")
+		}
+		_, flags := frm.OffsetAndFlags()
+		if flags != tcp.FlagSYN {
+			t.Fatalf("expected SYN packet, got %s", flags.String())
+		}
```
soypat: You ditched checking the second packet. Please try to maintain the equivalence with the original test

soypat (changes requested):

soypat:
Hey @MDr164, thanks for demonstrating how the synctest package can be used. I'm on the fence for using it since it seems it lacks the fine scheduling control one has with the ltesto.Sched type. You still rely on channels to synchronize the end of the test along with Wait. I also like the idea of getting comfortable with scheduling with channels since that seems to be the best path to integrating with the hardware and being performant in the process as in https://github.com/usbarmory/go-net/pull/2. This channel scheduling know how can then be pretty useful when designing an iouring styled interface. What are your thoughts on the channel scheduler approach in https://github.com/soypat/lneto/pull/143 ?

MDr164:
The PR you opened looks promising but I don't have the capacity to review it in much detail this week, the CI is happy which is a good sign already. Not using synctest also allows us to stay on go 1.24 so fine by me.

soypat:
I've merged the other PR, we can keep this PR open or create an issue to continue the discussion of synctest or improving the scheduler approach

MDr164:
Up to you, I'm fine with either. We can also close this one and simply discuss under #140 :)

soypat:
Discussion continues in #140

PR #143, opened 2026-06-25
use ltesto as package to contain scheduler/goroutine testing logic

soypat:


soypat:
Will read up on synctest to get an idea of how it works to better understand the pros/cons of each approach

PR #145, opened 2026-06-27
Netdev revamp

soypat:
`netdev` package should solve the ethernet interface and stack linking issue taking into consideration asynchronous functioning and polling functioning. This is unsurprisingly very hard to get right.


soypat:
@MDr164 As far as I can tell you were the only one using the netdev package so wanted to give you the heads up it has changed thoroughly. Specifically now netdev.Runner requires configuration and selecting the mode of operation. CYW43439 for example is Poll+Async mode where data is received via the async callback but which still requires some polling in the loop. I've also added the WakeOnRx mode which can enable higher throughput through faster response to incoming packets.

Note: unsafe package used to ensure 32bit aligned buffers for DMA on the cyw43439.

MDr164:
> @MDr164 As far as I can tell you were the only one using the netdev package so wanted to give you the heads up it has changed thoroughly.

Thanks! Yes I'm using it and I'll open source the code consuming it in the not so distant future as well. My code is also being moved around a lot recently so I don't mind adapting to an API change. I have one device where I used RX polling and another where I'm experimenting with DMA and interrupt based networking so I can try both. Will give it a review once I got back to the hardware on Monday.

MDr164:
i think there are some smaller issues with the buffers, I'll test to confirm my suspicion on my platforms if you don't need to merge it urgently.

soypat:
Please test throughly! I really want to get this one right. Will be trying to use it in espradio to replace their netlink package.

MDr164 (commented):
I found a couple subtle race(y) conditions. Overall it's an improvement over the old API in my opinion. I'll add the findings in more detail here throughout the day so please don't merge that one yet.

x/netdev/buffer.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+func (bs *bufferSelect) reset(bufs [][]byte) {
+	internal.SliceReuse(&bs.bufs, len(bufs))
+	bs.bufs = bs.bufs[:len(bufs)]
+	for i := range bs.bufs {
+		bs.bufs[i].buf = bufs[i]
+	}
+	bs.releaseAll()
```
MDr164, line 46: This only clears `lenAcquire` but not the `isRx` bool, when you now reuse the buffer via `SliceReuse` you still have this bool set to true it seems so `numFree` doesn't count right and sometimes makes the `EthPoll` call skip wrongly.

x/netdev/buffer.go
```diff
@@ -0,0 +1,96 @@
...
+func (bs *bufferSelect) forceAcquireTx(len int) (buf []byte, dropped int) {
+	for i := range bs.bufs {
+		if bs.bufs[i].lenAcquire.CompareAndSwap(0, int32(len)) {
+			return bs.bufs[i].buf, 0
+		}
+	}
+	bs.bufs[0].isRx.Store(false)
+	prevLen := bs.bufs[0].lenAcquire.Load()
+	bs.bufs[0].lenAcquire.Store(int32(len))
+	return bs.bufs[0].buf, int(prevLen)
```
MDr164: I think the CAS operation here is not quite right. Between `lenAcquire.Load()` and `Store()` a `recvEthHandler` on another goroutinecan run `CompareAndSwap(0, n)` successfully on slot 0 as both routines believe that they own the buffer. With the handler calling `copy` on incoming frames while the runner passes them to the egress frames which means partially constructed frames land on the wire eventually or it ends in the panic call. I think a proper fix is `CompareAndSwap(prevLen, int32(len))` which makes sure the whole thing is atomic.

x/netdev/runner.go
```diff
@@ -3,91 +3,261 @@ package netdev
...
+				case lneto.BackoffFlagGosched:
+					runtime.Gosched()
+					fallthrough
+				case lneto.BackoffFlagNop:
+					continue
+				default:
+					d = max(d, 100*time.Microsecond)
+				}
+				r.waketimer.Reset(d)
+				backoffs++
```
MDr164: I don't think this is actually called when your backoff is nop/gosched, is that intended?

MDr164 (commented):
I had a test myself and under most circumstances it works but it leaves a couple narrow race condition windows open. I wasn't able to pinpoint everything right myself as I also wasn't able to easily reproduce it so I had the clanker (Fable/Mythos) take a look and it generated this report:

<details>
<summary>Clanker Report</summary>

1. forceAcquireTx violates buffer ownership and corrupts in-flight Rx (buffer.go).
acquireNext correctly uses CAS on lenAcquire to claim a slot. forceAcquireTx does not — it falls back to plain Store on slot 0:
gobs.bufs[0].isRx.Store(false)
prevLen := bs.bufs[0].lenAcquire.Load()
bs.bufs[0].lenAcquire.Store(int32(len))
recvEthHandler runs on a different goroutine (driver goroutine on TinyGo, IRQ-woken goroutine on TamaGo). Interleaving: handler's CAS(0→n) on slot 0 succeeds between the runner's Load and Store; both parties now believe they own slot 0. The handler copy()s the incoming frame into the buffer while the runner hands the same buffer to EgressPackets and then SendOffsetEthFrame — a partially overwritten frame goes on the wire. Afterwards both call release, and the second one hits the panic("bs:race to release") — a production panic in the network loop, unrecoverable on TinyGo/TamaGo. The steal must itself be a CAS loop (CompareAndSwap(prevLen, len) and retry/give-up on failure), or the design should simply fail Tx for that iteration when no buffer is free instead of stealing.

2. Configure races against an in-flight receive handler (runner.go / interface contract).
Configure calls teardownAsync() (→ SetEthRecvHandler(nil)) and then bufs.reset(), which may reallocate bs.bufs via SliceReuse. Nothing in the DevEthernet contract says SetEthRecvHandler(nil) synchronizes with a handler invocation already executing on the driver goroutine. So recvEthHandler can read the bs.bufs slice header while reset swaps it — a torn 3-word slice-header read, i.e. memory unsafety, not just a logic race. The interface doc needs a quiescence guarantee (the synchronize_irq() semantic: "after SetEthRecvHandler(nil) returns, no handler invocation is in flight"), and drivers must be able to honor it. Same class of problem: Wake() reads r.flags and r.wake (plain fields) from arbitrary goroutines while Configure writes them under its own lock only.

3. service leaks the Tx buffer on the errEgressInvalidWrite path (runner.go).
return 0, 0, errEgressInvalidWrite happens after acquireNext/forceAcquireTx but before r.bufs.release(txbuf). The slot stays acquired forever. With the 1-buffer configuration in the picow example, the next iteration's forceAcquireTx steals it (incrementing pktlost spuriously); with larger pools it's a permanent slot leak per occurrence. Needs defer-style release or release-before-return.

4. Run rejecting EnableAsyncHandling also tears it down (runner.go).
The defer iface.dev.SetEthRecvHandler(nil) is installed before the r.asyncH != nil check. So calling Run in a configuration where it's forbidden (errAsyncHandlingWithRun) silently uninstalls the handler that EnableAsyncHandling set up — subsequent RunOnce calls then drop all async frames with no error. The check should precede the defer. Related: in poll-only mode the same defer clears whatever handler the user may have installed directly on the device, which Run never owned.

5. backoffs never increments on the Gosched/Nop continue path (runner.go).
In the wake branch:
gocase lneto.BackoffFlagGosched: runtime.Gosched(); fallthrough
case lneto.BackoffFlagNop: continue
continue skips backoffs++. A strategy that returns Gosched at consecutiveBackoffs == 0 (a normal opening move) pins the counter at 0 and busy-spins forever; the backoff never escalates to a timed sleep, and the select on r.wake is never reached. On TinyGo's cooperative scheduler a strategy returning Nop here without its own yield is a hard livelock. Increment before continue.


Design and correctness concerns:

Rx ordering. getRx scans slots in index order, so with >1 pending frame, ingress order is slot order, not arrival order. There's no generation counter. Reordering segments of the same TCP connection gratuitously triggers dup-ACKs/retransmit churn; DHCP/ARP exchanges are also order-sensitive in practice. A monotonic sequence stamped in acquireNext and honored by getRx fixes this cheaply.

acquireNext can panic on oversize frames. recvEthHandler calls acquireNext(len(incomingEthernet), true), and acquireNext slices buf[:len] without a bounds check. If a driver ever delivers a frame larger than the pool buffer, that's a slice-bounds panic inside the receive path. The picow example makes this plausible: MaxFrameSizeAndOffset was changed from cyw43439.MaxFrameSize to a hardcoded 2048 — if the driver constant and the magic number ever diverge, the pool is mis-sized. Bounds-check and drop (counting pktlost) instead.

Stale isRx after reset. reset clears lenAcquire via releaseAll but never clears isRx. When the backing array is reused, numFree() (!isRx && len==0) undercounts, which in async+poll mode can wrongly skip the EthPoll(nil) pump.

time.Timer Reset/Stop semantics. The wake loop Resets and Stops without draining waketimer.C. This is correct under Go 1.23+ synchronous-timer semantics but was wrong before, and TinyGo's runtime does not necessarily match upstream timer semantics version-for-version. Worst case here is a spurious wakeup (benign), but it deserves a comment stating the assumed semantics, given the target compilers.

Swallowed errors + println in the datapath. EthPoll, EgressPackets, and SendOffsetEthFrame errors are println'd and discarded. Two problems: (a) as a library, unconditional println in the hot loop goes to blocking UART on TamaGo/TinyGo — a persistent link-down condition turns the runner into a serial-port spammer that throttles the whole system; (b) send failures don't increment any counter, so pktlost (buffer exhaustion only) undercounts actual loss. These should route through an optional error callback or at minimum counters, and the poll branch currently drops the EthPoll error entirely whenever efrm > 0.

Single-frame batching asymmetry. Stack.EgressPackets/IngressPackets take [][]byte explicitly to batch, but the runner only ever passes bufsaux [1][]byte, and DevEthernet.EthPoll is one-frame-per-call. For CYW43439-class SPI devices that's fine; for TamaGo-class MACs with DMA descriptor rings (i.MX ENET, macb), per-frame calls through the interface forfeit exactly the batching the Stack API was shaped for. Not blocking, but the DevEthernet recv side should probably grow a batched form before drivers proliferate. Relatedly, RunnerBuffers' 4-byte alignment is fine for the current copy-based design but insufficient for future zero-copy DMA on A-class cores (cache-line alignment + maintenance); the doc comment "for Ethernet Rx buffer allocation" overpromises.

Stack.Socket returning any. Callers must type-assert to net.Conn/net.Listener, and the signature carries both the network string and family, sotype int — redundant encodings of the same thing. Splitting into Dial/Listen shaped like net.Dialer/net.ListenConfig would be more idiomatic and composes better with the TinyGo GoNet linkname shim in the same package (which is fd-based Berkeley sockets — a third socket vocabulary in one package; worth a doc note on which layer adapts to which). The commented-out Configure in the Stack interface should be removed or implemented.

Generic C parameter. Interface[C]/Runner[C] exist solely to thread Netlink reconnect params, and RunnerConfig.ReconnectParams is stored but never read anywhere in this PR — dead weight that infects every consumer's type signatures. Binding connect params at Netlink construction (closure/struct field) drops the type parameter entirely and simplifies both examples. If reconnection logic is planned for Runner, land it with that PR.

Doc drift. DevEthernet comments reference SendEthFrameOffset and SetRecvHandler; the methods are SendOffsetEthFrame and SetEthRecvHandler. The EthPoll dual contract ("writes to buf" vs. "buf is nil, pump only," keyed on whether a handler is set) is easy for driver authors to get wrong — a driver writing into the nil buf panics. Consider splitting into Poll() and RecvEthFrame(buf) or at least an explicit statement that the runner always passes nil in async mode.

Minor: RunOnce doesn't nil-check stack (→ panic in service) while Run does. forceAcquireTx counts a stolen stale Tx slot as pktlost. The espradio example still uses the removed 4-arg Run(ctx, iface, &stack, backoff) signature and passes iface by value — stale (underscore dir keeps it out of the build, but it ships broken).

Missing tests
x/netdev has zero test files. For a package whose entire value proposition is "get the async/poll linking right," the minimum bar: (a) a bufferSelect stress test — one goroutine hammering acquireNext(_, true) + release against another doing the Tx acquire/force/release cycle, run under -race on linux; this reproduces issues 1 and 2 deterministically enough. (b) A mock DevEthernet covering all four flag combinations, including the async+poll pump path and the wake path with a Gosched-returning backoff (catches issue 5). (c) A Run-after-EnableAsyncHandling test (catches issue 4). The RegisterUDP4 nil-raddr fix in stack-async.go looks correct (zero addr → wildcard remote) but also lands untested.

Bottom line
Architecture: sound, and notably better than the old single-mode runner for the CYW43439 reality of frames arriving during arbitrary bus servicing. Issues 1–5 are blocking; 1 and 2 are memory-safety-adjacent and will be intermittent, hardware-timing-dependent failures in the field — exactly the class of bug this revamp exists to eliminate. With a CAS-correct steal path (or no steal), a quiescence contract on SetEthRecvHandler, the leak/teardown/increment fixes, and a race-detector-backed test for the pool, this is mergeable.
</details>

x/netdev/runner.go
soypat: That's because the waketimer needs to be reset only if hitting the select statement. otherwise it is a ineffective call that can be skipped.

x/netdev/buffer.go
soypat: OK, I've made a hefty rewrite with the help of claude but carefully reviewed it myself. Should be taken care of with lenClaim flag value for lenAcquire

x/netdev/buffer.go
soypat: Alright, should be taken care of in latest commit

MDr164:
> @bench-report/bench-report.md

Ok it should comment with the content not the filename. I'll investigate that later but shouldn't bother much for now.

I will review the changes in a bit and assume the tests are fine (I get reading fatigue fast when looking at 600+ lines of testing code)

x/netdev/buffer.go
MDr164: Looks solid now

x/netdev/buffer.go
MDr164: Looks good :+1:

x/netdev/runner.go
```diff
@@ -3,91 +3,328 @@ package netdev
...
+				switch d {
+				case lneto.BackoffFlagGosched:
+					runtime.Gosched()
+					fallthrough
+				case lneto.BackoffFlagNop:
+					continue
+				default:
+					d = max(d, 100*time.Microsecond)
+				}
+				// Claude say:
```
MDr164, line 312: nitpick: should this line stay in the code? I don't mind really, just making sure it wasn't forgotten here :)

x/netdev/buffer.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"sync/atomic"
+
+	"github.com/soypat/lneto/internal"
+)
+
+// TODO(soypat): True Zero Copy (TZC)
+// TODO(soypat): TZC acheived on redesigning [DevEthernet] to not own any buffers and ask the networking stack for buffers in the callback path. TZC already acheived for Polling path.
+// TODO(soypat): TZC in callback path requires redesign of bufferSelect and [Runner] likely.
+
```
MDr164, line 12: My clanker had one more TODO item which I personally didn't catch and also didn't run into in practice, so no idea how likely that problem is:

> With the steal removed, the fallback is: drain up to len(bufs) frames, retry acquire, and skip Tx if it still fails. Under an rx flood the handler races the runner for every freed slot; on preemptive multicore builds Tx wins only statistically, and in async+poll mode the drain loop's EthPoll(nil) pump can refill slots as fast as it drains them, so the for range len(bufs) bound doesn't guarantee convergence. TCP self-throttles via the unACKed window, but ARP/ICMP/ND replies can starve outright. A dedicated Tx slot (reserve slot 0, or a separate txbuf outside the pool) makes Tx progress deterministic and costs one buffer. Not blocking — the skip is safe, unlike the steal — but it's the main remaining design gap and worth a TODO alongside the TZC ones.

x/netdev/runner.go
```diff
@@ -96,20 +333,68 @@ func (r *Runner[C]) Run(ctx context.Context, iface Interface[C], stack Stack, ba
 	return ctx.Err()
 }
 
+func (r *Runner[C]) service(iface *Interface[C], stack Stack, bufsize int, poll, async bool) (nrx, ntx int, err error) {
+	nrx, err = r.doRx(iface, stack, poll, async)
```
MDr164, line 337: The err here is not being used it seems

x/netdev/runner.go
soypat: Yeah, figured to leave it in since I couldnt tell if it was 100% valid comment.

x/netdev/buffer.go
soypat: This is actually true and something I did worry about. I also thought of the solution where you always have 1 buffer for tx but that breaks down when you want to share 1 for tx and rx. Maybe a credit system could solve this issue. Will look into it

x/netdev/runner.go
soypat: Yeah, did also notice that and was about to remove the error but on second thought will give it another think through

MDr164:
Had a more detailed look myself and also had Fable/Mythos re-review and there is nothing critical left in this PR and I'm testing it on my rv64 board over a longer time now but so far looks solid. So just the unchecked error and the potential TX starve left. GitHub doesn't allow me to mark the other conversations as resolved.

soypat:
ahhh merging for now. will create issues with what you noted

PR #146, opened 2026-06-27
Add test to guarantee IPv6 not included in xnet.StackAsync accidentally

soypat:
Also rename the test file to be more clear: it is untinygo, rather than tinygo oriented test file.


MDr164:
This looks pretty similar to the test we had in u-root to check the same, the test was pretty valuable there so seems like a good idea for this project as well :+1:

soypat:
Alright! Merged it is!

Issue #144, opened 2026-06-27
IPv6 support tracking issue


MDr164:
This issue will serve as an umbrella issue for tracking the gradual upstreaming of the IPv6 work that currently lives on a development branch. There is a large amount of code spanning addressing, extension headers, ICMPv6/NDP, DHCPv6 (client and server) and the `x/xnet` stack integration, so rather than land it as one change it will be split into smaller, independently reviewable PRs. Each checkbox below is intended to map to roughly one PR. I'll gradually refine the bullet points as work is upstreamed.

- [ ] Add IPv6 scoped address representation
    - Add `ipv6.ScopedAddr` with a numeric zone identifier and `NewScopedAddr` constructor (RFC 4007).
    - Add `AppendFormatScopedAddr` and `HasZone`/`Scope` accessors.
    - Cover zone formatting, link-local zoning, and round-trip parsing in tests.
- [ ] Add IPv6 address scope classification
    - Add `ipv6.Scope`, `AddrScope`, `IsMulticast`, and `IsLinkLocalUnicast` helpers (RFC 4291/4007).
    - Define the scope constants and document the precedence between unicast/multicast scopes.
    - Add table-driven tests for global, link-local, unique-local, loopback, and multicast scopes.
- [ ] Add RFC 6724 default address selection
    - Add `SelectSourceAddr` with `SourceCandidate` and `preferSource`/`addrPrecedence`/`addrLabel` rules.
    - Add `SortDestinationAddrs` with `DestinationCandidate` and `preferDestination` ordering.
    - Add `commonPrefixLen` and `isIPv4Mapped` helpers and test the policy table against the RFC examples.
- [ ] Add SLAAC address derivation
    - Add `SLAACAddrFromMAC` for stable modified EUI-64 addresses (RFC 4862).
    - Add `TemporarySLAACAddr` for privacy/temporary addresses (RFC 8981).
    - Test prefix-length validation, the universal/local bit flip, and deterministic temporary generation.
    - Decide scope for the address lifecycle (DAD, preferred/valid lifetimes) — out of scope here or a follow-up.
- [ ] Add IPv6 extension header chain walking
    - Add `ipv6.ExtensionHeader`, `Frame.ForEachExtensionHeader`, `isExtensionHeader`, and `extensionHeaderLength` (RFC 8200).
    - Return the upper-layer protocol and payload offset after the chain.
[...]

PR #147, opened 2026-06-29
Handle IA_NA/IA_PD requests and config options in DHCPv6

MDr164:
Adds the DHCPv6 client request exchange with IA_NA/IA_PD handling, reconfigure-renew (RFC 8415), DNS server and domain search options (RFC 3646) and NTP server suboptions (RFC 5908).

### DIFF
--- a/dhcp/dhcpv6/client.go
+++ b/dhcp/dhcpv6/client.go
@@ -6,6 +6,7 @@ import (
 	"net/netip"
 
 	"github.com/soypat/lneto"
+	"github.com/soypat/lneto/dns"
 	"github.com/soypat/lneto/internal"
 )
 
@@ -14,6 +15,13 @@ type RequestConfig struct {
 	// ClientHardwareAddr is the client's Ethernet MAC address.
 	// It is used to construct the client DUID-LL and IAID.
 	ClientHardwareAddr [6]byte
+}
+
+// DelegatedPrefix is an IPv6 prefix delegated by a DHCPv6 server.
+type DelegatedPrefix struct {
+	Prefix            netip.Prefix
+	PreferredLifetime uint32
+	ValidLifetime     uint32
 }
 
 // Client is a stateful DHCPv6 client implementing the [lneto.StackNode] interface.
@@ -39,6 +47,21 @@ type Client struct {
 	// dns accumulates DNS recursive name server addresses (OptDNSServers).
 	// Client owns the backing array; cleared on reset, capacity reused.
 	dns []netip.Addr
+	// domainSearch accumulates DNS domain search names (OptDomainList).
+	// Client owns the backing array; cleared on reset, capacity reused.
+	domainSearch []dns.Name
+	// ntps accumulates NTP server addresses (OptNTPServer, suboption 1).
+	// Client owns the backing array; cleared on reset, capacity reused.
+	ntps []netip.Addr
+	// ntpMulticast accumulates NTP multicast addresses (OptNTPServer, suboption 2).
+	// Client owns the backing array; cleared on reset, capacity reused.
+	ntpMulticast []netip.Addr
+	// ntpNames accumulates NTP server FQDNs (OptNTPServer, suboption 3).
+	// Client owns the backing array; cleared on reset, capacity reused.
+	ntpNames []dns.Name
+	// delegatedPrefixes accumulates IA_PD prefixes (OptIAPD/OptIAPrefix).
+	// Client owns the backing array; cleared on reset, capacity reused.
+	delegatedPrefixes []DelegatedPrefix
 
 	assignedAddr      [16]byte
 	assignedAddrValid bool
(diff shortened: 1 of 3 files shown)


MDr164:
First chunk of the IPv6 additions mentioned in #144 which adds a couple of items from that list but they are kind of codependent so I published it in one PR. From experiments I used a couple of numbers directly in the code instead of defining constants. I think other than the missing constants it's reviewable.

dhcp/dhcpv6/client.go
```diff
@@ -221,18 +298,120 @@ func (c *Client) setOptions(frm Frame) error {
...
+		subLen := int(binary.BigEndian.Uint16(data[ptr+2:]))
+		if ptr+4+subLen > len(data) {
+			break
+		}
+		if subCode == OptIAPrefix && subLen >= 25 {
+			sub := data[ptr+4 : ptr+4+subLen]
+			bits := int(sub[8])
+			prefix := netip.PrefixFrom(netip.AddrFrom16([16]byte(sub[9:25])), bits)
+			if prefix.IsValid() {
+				c.delegatedPrefixes = append(c.delegatedPrefixes, DelegatedPrefix{
```
soypat, line 434: ungated alloc

dhcp/dhcpv6/client.go
```diff
@@ -221,18 +298,120 @@ func (c *Client) setOptions(frm Frame) error {
...
+		if ptr+4+subLen > len(data) {
+			break
+		}
+		subData := data[ptr+4 : ptr+4+subLen]
+		switch subCode {
+		case 1:
+			if subLen != 16 {
+				break
+			}
+			c.ntps = append(c.ntps, netip.AddrFrom16([16]byte(data[ptr+4:ptr+20])))
```
soypat: ungated alloca

dhcp/dhcpv6/client.go
```diff
@@ -221,18 +298,120 @@ func (c *Client) setOptions(frm Frame) error {
...
+		case 2:
+			if subLen != 16 {
+				break
+			}
+			c.ntpMulticast = append(c.ntpMulticast, netip.AddrFrom16([16]byte(data[ptr+4:ptr+20])))
```
soypat: ungated alloca

dhcp/dhcpv6/client.go
```diff
@@ -221,18 +298,120 @@ func (c *Client) setOptions(frm Frame) error {
...
+		case 3:
+			idx := len(c.ntpNames)
+			if idx < cap(c.ntpNames) {
+				c.ntpNames = c.ntpNames[:idx+1]
+			} else {
+				c.ntpNames = append(c.ntpNames, dns.Name{})
```
soypat, line 477: allocs more when needed without gate

dhcp/dhcpv6/client.go
```diff
@@ -221,18 +298,120 @@ func (c *Client) setOptions(frm Frame) error {
...
+func (c *Client) parseDomainSearch(data []byte) {
+	if len(c.domainSearch) > 0 {
+		return
+	}
+	for off := uint16(0); off < uint16(len(data)); {
+		idx := len(c.domainSearch)
+		if idx < cap(c.domainSearch) {
+			c.domainSearch = c.domainSearch[:idx+1]
+		} else {
+			c.domainSearch = append(c.domainSearch, dns.Name{})
```
soypat, line 498: more alloca

soypat (commented):
My first observation is that it looks good- however there are several ungated allocations. Do you think it is worth adding a couple more configuration parameters (max ntpnames, max delegatedPrefices) so as to prevent OOM attacks? after we address this I'll look more deeply into the PR

MDr164:
Quick question, probably not worth opening an issue for: Should I follow through and potentially extend a functional options pattern here and the rest of the codebase more? E.g. the WithXXX(option) stuff so that we always get sane default but can modify more values as needed

soypat:
What case are you looking at? I have not found functional options to give benefits worth paying extra heap allocations for in lneto's particular situation.

MDr164:
Ok nevermind scrap that idea, I forgot about the allocations

dhcp/dhcpv6/client.go
MDr164: Now gated by MaxDomainSearch

dhcp/dhcpv6/client.go
MDr164: Now gated by MaxNTPServerNames

dhcp/dhcpv6/client.go
MDr164: Now gated by MaxNTPMulticastServers

dhcp/dhcpv6/client.go
MDr164: Now gated by MaxNTPServers

dhcp/dhcpv6/client.go
MDr164: Now gated by MaxDelegatedPrefixes

MDr164:
@soypat should be ready to review now, 60% of this PR are tests again. I noticed that the last PRs I published are quite large but more than half is usually tests so I guess that's fine? I'll try to make them a bit smaller in the future.

soypat (approved):
Alright, looks good!

PR #148, opened 2026-06-30
Add out-of-order segment reassembly

MDr164:
This is part 1 of the split for #115 

Add an opt-in, bounded out-of-order reassembly buffer so a single lost segment can be recovered by retransmitting the gap while later segments are held and delivered once the gap fills.

The receiver also subtracts buffered out-of-order bytes from the advertised receive window and avoids challenge-ACK aborts for in-window future data. Reassembly is disabled by default.

Done with clanker assistance.

### DIFF
--- a/tcp/handler.go
+++ b/tcp/handler.go
@@ -28,12 +28,38 @@ type Handler struct {
 	// connection is established via Open calls. This disambiguates whether
 	// Read and Write calls belong to the current connection.
 
-	optcodec   OptionCodec
+	optcodec OptionCodec
+	// reasm is the optional out-of-order reassembly buffer. When disabled
+	// (no buffer set) the Handler only accepts in-order segments, as before.
+	// See [Handler.SetReassemblyBuffer].
+	reasm      reassembly
 	closing    bool
 	shutdownRx bool
 	// nRetransmit stores the number of times the oldest packet was retransmit.
 	nRetransmit    uint8
 	requeueControl bool
+}
+
+// SetReassemblyBuffer enables out-of-order segment reassembly using buf as
+// storage for up to maxSegments segments that arrive ahead of the next expected
+// sequence number (buf is divided into maxSegments equal slots). With
+// reassembly enabled a single lost segment is recovered by retransmitting only
+// the gap; the buffered tail is then delivered without go-back-N. Passing a nil
+// buffer or zero maxSegments disables reassembly (the default: in-order only).
+// Returns [lneto.ErrBadState] if the connection is open.
+//
+// buf should be sized at least as large as the receive buffer so that any
+// in-window segment arriving during a gap can be held: the receiver subtracts
+// out-of-order bytes from its advertised window, but the slab must be able to
+// absorb a full window's worth of out-of-order data. Each slot holds one
+// segment, so maxSegments bounds how many distinct out-of-order segments may be
+// outstanding; len(buf)/maxSegments must be at least one segment (MSS) of bytes.
+func (h *Handler) SetReassemblyBuffer(buf []byte, maxSegments int) error {
+	if !h.scb.State().IsClosed() {
+		return lneto.ErrBadState
+	}
+	h.reasm.reset(buf, maxSegments)
+	return nil
 }
 
 // SetLoggers sets the [slog.Logger] for the Handler and internal [ControlBlock].
@@ -131,9 +157,11 @@ func (h *Handler) reset(localPort, remotePort uint16, iss Value) {
 		remotePort: remotePort,
 		validator:  h.validator,
 		logger:     h.logger,
+		reasm:      h.reasm,
 		closing:    false,
 		shutdownRx: false,
 	}
+	h.reasm.clear() // preserve the slab/config across reopen, drop held segments.
 	h.bufTx.ResetOrReuse(nil, 0, iss)
 	h.bufRx.Reset()
 }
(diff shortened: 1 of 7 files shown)


soypat:
good stuff- however I had something else in mind. Leveraging the existing Rx buffer for reassembly seems to me like a natural path forward- after all any data received should fit in this buffer given the normally advertised window (and if CC were to advertise a larger window it can do its own reassembly and hand off data to handler when the handler has space in it's buffer). What do you think?

MDr164:
Sounds ok to me, I can test if that works easily. This was more of a lazy approach that didn't require too much thinking on my side that still works decently well :sweat_smile:

soypat:
always ping me if you want something reviewed! Else I might gloss over thinking you are still working on it :)

MDr164:
@soypat noted, I think it's good to go now. I double checked myself that everything is coherent now. I can strip some of the comments especially if you don't like the somewhat obvious if/else condition ones but I kept loosing track which situation was dealing with what. The tests are fully clanker speak as I got no actual network here where ooo packets are a real issue currently.

tcp/handler.go
```diff
@@ -163,15 +169,22 @@ func (h *Handler) Recv(incomingPacket []byte) error {
 		return lneto.ErrMismatch
 	}
 	payload := tfrm.Payload()
-	if !h.shutdownRx && len(payload) > h.bufRx.Free() {
-		return lneto.ErrBufferFull
-	}
```
soypat, line 168: Hol' up. if payload is larger than Free then we have an issue, whether we can reassemble or not. We use free space for reassembly. I feel like this check should live here.

tcp/handler.go
```diff
@@ -204,6 +217,11 @@ func (h *Handler) Recv(incomingPacket []byte) error {
 			return err
 		}
 	}
+	if segIncoming.DATALEN != 0 && h.reasm.buffered() > 0 {
+		// The just-accepted in-order segment may have filled a gap; deliver any
+		// now-contiguous buffered segments.
+		h.flushReassembly()
```
soypat: isn't this the actual reassembly part to name it like so: `h.reassemble()` or `h.tryReassemble()`

tcp/reassembly.go
```diff
@@ -0,0 +1,117 @@
...
+	for i := 0; i < len(r.held); {
+		if r.held[i].seq.LessThan(nxt) {
+			r.held = append(r.held[:i], r.held[i+1:]...)
+			n++
+			continue
+		}
+		i++
+	}
+	return n
+}
```
soypat: using eager overwrite here makes sense if we keep held in order of seq
```go
func prune(nxt)
  i=0
  for i in held
    if held[i].seq < nxt || held[i].n==0
       continue
    else
       break
if i != 0
   held = append(held[:0], held[i:])

```

tcp/reassembly.go
```diff
@@ -0,0 +1,117 @@
...
+			return true // already buffered; idempotent.
+		}
+		heldEnd := Add(r.held[i].seq, Size(r.held[i].n))
+		if seq.LessThan(heldEnd) && r.held[i].seq.LessThan(end) {
+			return false // overlaps a held segment.
+		}
+	}
+	if len(r.held) >= cap(r.held) {
+		return false // metadata full.
+	}
```
soypat: this should be an early check

tcp/reassembly.go
```diff
@@ -0,0 +1,117 @@
...
+	gap := int(Sizeof(rcvNxt, seq))
+	if !rx.PeekWrite(payload, gap) {
+		return false // does not fit in the free region.
+	}
```
soypat, line 99: worth an early free space check too?

tcp/handler.go
```diff
@@ -242,6 +260,62 @@ func (h *Handler) Recv(incomingPacket []byte) error {
...
+func (h *Handler) flushReassembly() {
+	if h.shutdownRx {
+		// Discard mode skips the gap-filling write, so staged bytes can no
+		// longer be committed coherently; drop them (the peer retransmits).
+		h.reasm.clear()
+		return
+	}
+	for h.reasm.buffered() > 0 {
+		nxt := h.scb.RecvNext()
+		h.reasm.prune(nxt) // Drop stale/overlapping segments before delivering.
```
soypat: this will thrash the cpu since prune is looping and copying inside of another loop.

tcp/handler.go
```diff
@@ -242,6 +260,62 @@ func (h *Handler) Recv(incomingPacket []byte) error {
...
+		seg, ok := h.reasm.popContiguous(nxt)
+		if !ok {
+			break // a gap remains before the next buffered segment.
+		}
+		if err := h.bufRx.Commit(seg.n); err != nil {
+			break // ring lacks room; deliver the remainder on a later Recv/Read.
+		}
+		h.scb.rcv.NXT.UpdateForward(Size(seg.n))
+		h.scb.pending[0] |= FlagACK
```
soypat, line 307: gotta reimagine this function. this does not look very good. i feel there is a leaked abstraction. consider a method on reassembly and reducing the inputs. 

```go
func (*reassembly) reassemble(*internal.Ring, currentNXT Value, ???)

```

tcp/reassembly.go
```diff
@@ -0,0 +1,117 @@
...
+	r.held = append(r.held, reasmSeg{seq: seq, n: len(payload)})
```
soypat: this will always append a segment at the end and possibly make the seq value be out of order. Do we want to keep this list ordered? It will give us great simplicity benefits when operating on it. I think it is worth adding in-order and documenting held is ordered by oldest to newest seq.

tcp/reassembly.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return n
+}
+
+// store stages payload at the offset it will occupy in rx once the gap from
+// rcvNxt fills. It returns true when held, including when already held (storing
+// is idempotent). It fails if reassembly is disabled, the payload is empty,
+// metadata is full, it does not fit rx's free region, or it overlaps a held
+// segment.
+func (r *reassembly) store(rx *internal.Ring, rcvNxt, seq Value, payload []byte) bool {
```
soypat, line 69: If we keep held in order i feel like this function can be simplified too, specifically for loop contents.

soypat (commented):
have not looked at tests. a few suggestions

tcp/handler.go
MDr164: Yes, I wasn't too sure about naming. To me it was more of a flush collected packets and reassemble kinda flow hence that name. I can also just go for `reassemble()` instead.

tcp/handler.go
MDr164: I think I should double check my code on a microcontroller as my emulation and actual hardware didn't really make the problem show up but I do see what you mean and will rethink this.

tcp/handler.go
MDr164: This function also gave me the biggest headache so far, that's also why there are all these inline comments to keep track of what's happening where in the loop. I'll revisit this then.

tcp/reassembly.go
MDr164: I can add more checks, sure. I thought this extra check would not be needed here as the space check is kinda part of the rings peek method.

soypat (approved):

soypat:
looks pretty good, let's get the ball rolling on this :)

PR #149, opened 2026-06-30
Add benchci for better visualization

MDr164:
This adds a small in-repo tool and updates the ci.yml file in order to make benchmark and most importantly allocation results visible in a PR. It also drops the 3rdparty action we had commented out. For now this is just a CI UX test.

### DIFF
--- a/.github/workflows/bench-comment.yaml
+++ b/.github/workflows/bench-comment.yaml
@@ -1,0 +1,78 @@
+name: Benchmark Comment
+
+# Rationale: This more privileged workflow runs in the base repo's
+# context (with a write token) only after the 'untrusted' CI workflow finishes.
+# It does NOT check out or execute PR code; it only consumes the benchmark
+# report artifact as inert, validated data.
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+
+permissions:
+  contents: read
+
+jobs:
+  comment:
+    # Only for PR-triggered CI runs that succeeded to suppress noise
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write # needed to comment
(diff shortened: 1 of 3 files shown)


MDr164:
Generated a table here: https://github.com/soypat/lneto/actions/runs/28445447588/attempts/1#summary-84293652944

We can't reliably do comments like the codecov ones from forks like mine, only PRs from the same repo here get an autocomment with the current permissions (for better or worse). I'll experiment more to get someting more obvious than the somewhat hidden job summary

MDr164:
Hm, got another idea, let me try something with a second yaml file that separates permissions a bit more cleanly

soypat:
If this works reliably this would be hugely valuable. I'm worried the compute given to CI is mushy and may give us false impressions on days we are given more/less compute.

MDr164:
I'm just experimenting a bit for now and mainly testing the memory benchmarks which should be consistent across runs even if the CI environment schedules differently. For CPU resources there might of course be variance.

soypat:
Ah good point, allocations can be reliably tracked. We can mark the CPU usage as being unreliably in the table noting the main use is to track allocs. In that case I'm all for it!

MDr164:
Ok looks like the new yaml file is only being picked up once it is merged, that's not ideal for testing, GitHub, but understandable to be fair...

I'll play around with the report format a bit more but I think the YAML split should work and does not require any external actions and simply runs a shell script with the CI runners tools.

MDr164:
Imo the change is ready now, not my proudest work but feels like an upgrade. The newly added yml is purely for posting comments like the codecov one and consumes the markdown file which is already being rendered in the somewhat hidden run summary. We can also not include that one, up to you. The go bench to markdown tool is also rather crude and could likely be improved but I also didn't want to overengineer it tbh

MDr164:
Btw I created a githook I use locally which also checks the code for allocations before I can push, want me to add that as well or is there no need for that?

soypat:
Add the githook in a different PR- I am very interested in it!

PR #153, opened 2026-07-07
Make github action comment work right

MDr164:
It was previously just commenting the file name instead of the content, this change attempts to fix it using json instead of the raw file.

### DIFF
--- a/.github/workflows/bench-comment.yaml
+++ b/.github/workflows/bench-comment.yaml
@@ -35,7 +35,10 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.workflow_run.pull_requests[0].number }}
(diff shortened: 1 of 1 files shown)


MDr164:
I think I actually implemented myself into a corner here. The good thing: It's done in way to prevent a malicious PR to extract or change repo data as I made it workflow_run instead of workflow_dispatch so it only runs from main and not the PR branch. The bad thing: I can't test it here :|

.github/workflows/bench-comment.yaml
```diff
@@ -55,11 +58,17 @@ jobs:
             exit 1
           fi
 
-          # Resolve the PR from the run's head SHA rather than trusting any
-          # value carried in the artifact, so we can only comment on the PR that
-          # actually produced this run.
```
soypat, line 60: could you restore comments (if they are meaningful)

MDr164: Updated it, the old comment was not fully true anymore with the change as it sometimes just didn't publish a comment as it didn't find the PR via the SHA.

soypat (approved):

PR #155, opened 2026-07-10
TCP congestion part 2: Add RFC 6298 RTO estimator state machine

MDr164:
This is the next split of the TCP-CC PR #115  and adds a self-contained, allocation-free round-trip-time estimator and single retransmission timer following RFC 6298: SRTT/RTTVAR smoothing via integer shifts, RTO clamping, Karn's algorithm, and exponential backoff on timeout.


soypat:
Great job on the benchmark bot! Does it update the comment on every change?

tcp/control.go
```diff
@@ -79,8 +80,29 @@ type ControlBlock struct {
...
+	// timer. When nil, time.Now is used. Set via SetClock for deterministic tests.
+	clock func() time.Time
+}
+
+// now returns the current time from the injected clock, or time.Now if unset.
+func (tcb *ControlBlock) now() time.Time {
+	if tcb.clock != nil {
+		return tcb.clock()
+	}
+	return time.Now()
```
soypat: woah woah woah. The RTO come separately, remember the interface approach we were originally going for? Revisit our comment exchanges from a few weeks ago

soypat: we want timing to be outside the TCP package

MDr164: Yes, that's why it's still a draft. I simply opened the PR from the split out code without any changes so far. The TCP-CC PR existed a bit longer than the new RTO code we recently did so this is partially stale code.

MDr164:
> Great job on the benchmark bot! Does it update the comment on every change?

It should, unless I got the gh cli calls in the script wrong. We'll see.

tcp/control.go
MDr164: @soypat I moved the `time.Now()` call out to adapt to the last discussions we had in the other PR. The import to `time` is still there for the function signature which has `time.Duration` but I could also change that one to uint or so probably if needed. Should be good to review now.

tcp/control.go
```diff
@@ -79,7 +80,29 @@ type ControlBlock struct {
...
+	// rto is the RFC 6298 round-trip-time estimator and single retransmission
+	// timer. Timing lives outside the tcp package (see the tcp/rto leaf package):
+	// it is fed by Send (timing new data) and Recv (sampling ACKs) and consulted
+	// by CheckRetransmitTimeout, but only while a clock has been injected.
+	rto rto.Control
+	// clock injects the current monotonic time in nanoseconds (the func() int64
+	// convention used across lneto). Time integration is opt-in: when clock is
+	// nil the retransmission timer stays dormant and the tcp package performs no
+	// timing at all. Set via SetClock.
+	clock func() int64
```
soypat: but we said no clock in tcp package (a.k.a no timing dependency in tcp package).
clock should be part of the rto.Control if anything.

soypat (changes requested):
Sorry, i saw the clock field and could not go any further.

rto.Control can drive outgoing packets from an early check in tcp.Conn.Encapsulate. I don't see the reason to incorporate this field outside of the interface that drives timing. Let's keep tcp package free from timing which is a very complex subject which would draw many opinions- which I'd rather drag far away from the cold-hard sequence numbers we know and love to be written in stone.

soypat:
I want to discuss the future of lneto before continuing here! https://github.com/soypat/lneto/discussions/157

MDr164:
I added 3 commits on top which deal with the outcome of the discussion. I also added a bunch of pointer to said discussion in the code comments. Let me know if you want those not part of the code (I was rather conflicted with leaving git platform dependent comments in). I'm giving this another round of local testing but I think it's fine for an initial look.

tcp/control.go
```diff
@@ -662,11 +662,15 @@ func (tcb *ControlBlock) rstJump() Value {
...
-// Implements RFC 9293 §3.10.8 (RETRANSMISSION TIMEOUT).
-// func (tcb *ControlBlock) Retransmit() { tcb.snd.NXT = tcb.snd.UNA }
+// Retransmit rewinds snd.NXT back to snd.UNA so the next PendingSegment and
+// Send calls retransmit all unacknowledged data from the oldest sequence number
+// (go-back-N). It must be paired with ringTx.RetransmitFromUNA to rewind the
+// transmit buffer. Implements RFC 9293 §3.10.8 (RETRANSMISSION TIMEOUT).
+func (tcb *ControlBlock) Retransmit() {
+	tcb.snd.NXT = tcb.snd.UNA
+	tcb.dupack = 0
+	tcb.nRetransmit = 0
```
soypat, line 672: This unchecked would lead to an inconditional retransmission of ALL previously sent data. Were we not adding reassembly to prevent this?

tcp/handler.go
```diff
@@ -337,6 +376,18 @@ func (h *Handler) Send(b []byte) (int, error) {
...
+	var now int64
+	if h.loss != nil {
+		now = h.nanotime()
+		if h.loss.PreTx(now).Retransmit {
+			// Go-back-N retransmission directed by loss recovery: rewind the
+			// send sequence and transmit buffer so unacknowledged data is resent
+			// from snd.UNA. Done before the early short-circuit below so an
+			// expired RTO retransmits even with no new data queued.
+			h.scb.Retransmit()
+			h.bufTx.RetransmitFromUNA()
```
soypat, line 388: So for now we only do Retransmit (only use one of TxDirective fields), right? The plan would be to add more complexity later on? Also I noticed LossRecovery would need to duplicate the TCB to know what our state is, which if RTO does not need that's fine, we can modify LossRecovery in the future when working with CongestionControl. We can go the learning way. I like we started out dead simple

soypat (commented):
Looks tight. I'm liking the direction, few open ended questions, but no real blocker to merging this as is to start experimenting and learning what works, what does not. I'm open to reworking this interface in the future (I've got a feeling we will :) )

tcp/control.go
MDr164: iirc reassembly was just for re-ordering packets that have been received in the wrong order but I'm also juggling a bunch of projects so maybe I'm remembering wrong, will have a look again.

tcp/handler.go
MDr164: Gimme a second, I got like 3 more local branches for the whole tcp-cc patch train. I might have those changes you mention here already in one of those. Just need to dig through the chaos here.

tcp/control.go
MDr164: Selective retransmission is in https://github.com/MDr164/lneto/commits/split/04-sack/ and the reassembly stuff is receiver-side only so I did actually remember correctly.

tcp/handler.go
MDr164: https://github.com/MDr164/lneto/commits/split/04-sack/ and the 05 and 06 split branches (the last two I still need to rebase) fill the remaining gaps you mentioned. If you'd like I can already post all split PRs as drafts to visualize the diff better.

soypat: Whatever makes the diff easier to read is A-OK by me. I've understood this PR's diff pretty well so keep that up

MDr164:
@soypat I uploaded part 3 and part 4 of the TCP-CC split. I think they cover the missing pieces you were looking for if I'm not mistaken. That said this current PR remains unchanged for now as I'm adding the last parts in the next two parts and then the final actual congestion interface itself in part 5. Part 6 adds the CUBIC/BBR examples and conclude the whole TCP reliability patch train from my side.

EDIT: currently rebaseing the other split PRs and will publish the whole patch train I got locally. Each part contains the commits from the previous parts so you might need to look at individual commits instead of the whole PRs.

EDIT2: If this feature was already rolled out I would have been able to better organize the flood of incremental PRs a bit more neatly: https://github.github.com/gh-stack/

tcp/control.go
soypat: This is not what I had in mind for an easier to review change. If I understand correctly you are including Retransmission logic to then plug into 04-sack branch? I don't like the idea of juggling changes across PRs if we can avoid it, and it seems we can. Just stick to the Loss interface and build from first principles please. No juggling of requirements between PRs.

soypat:
The existence of `N` new PRs and the implication of dependency between them makes this much harder to review. I don't appreciate this approach. The road to Loss interface can be massively simplified. Drop `rto` and just implement the interface at the tcp package level.  You can add a simplified Loss interface implementation in tests to ensure it works. This seems like a much simpler path to reviewing and ensuring correctness. It also leads to the same end result with much lower cognitive load. Lets do this instead. Sorry I'm backtracking, I had no idea what you were going to come up with before you sent me these PRs.

soypat:
If the path is not clear let me know, I have no issue creating a PR to illustrate the path I'm imagining, after which you can take on the torch if you so desire :)

MDr164:
Well not all of them actually rely on each other, that's just the order I had split them all up from the super large original tcp-cc branch and to not deal with making them all standalone I simply added commits in order that I split it up hence the current dependency. At least part 5 and 7 are fully standalone iirc, part 3 is super small as well and another RFC altogether. I can merge the rto stuff back to the tcp package instead of a subpackage but it doesn't really change that much in regards of the open PRs. It's essentially just the work that piled up so I'd rather have it posted than randomly hanging around on my fork :)

MDr164:
I'll quickly edit the PRs to point to different branches so that you can see their diff better, for now only part 2 to 5 are really relevant for now

EDIT: Nevermind, GitHub won't allow me to

MDr164:
@soypat so I thought about how to change the code in this PR to align with some of the simplifications you requested. I might need a little input on how to proceed as I currently see the options:

Target shape:
- LossRecovery, TxDirective, RxDirective, RTTObserver — stay defined in tcp (unchanged).
- rto.Control (the RFC 6298 machine) moves into tcp as the default in-package implementation — e.g. tcp/loss_rto.go with an unexported rtoControl plus a small exported constructor, or keep it exported if introspection (SmoothedRTT/CurrentRTO) is wanted publicly.
- ConnConfig.LossRecovery + Nanotime + the "LossRecovery requires Nanotime" validation — stay.
- Testing becomes two clean layers:
- a trivial fake LossRecovery in tcp tests (deadline + "retransmit when now ≥ deadline") that proves the Handler drives the hooks and honors directives — this is the maintainer's "simplified implementation to ensure it works";
- the existing rto_test.go content, moved in-package, that unit-tests the RFC 6298 math.

That leaves following questions:
1. Interface visibility of the default impl: keep the RTO type exported (so users/tests can introspect SmoothedRTT/CurrentRTO) or unexported with the interface as the only public surface?
2. Scope of the this PR here: do you want the full RFC 6298 RTO shipped in this PR, or literally just the interface + the fake, with the real RTO following as its own PR afterwards?

It does not really change a lot in regards of the other parts. I can simply close those PRs again and re-open once previous/standalone PRs are merged (or not).

soypat:
Was easier to demo with code #168

MDr164:
With #168 now merged I moved the remaining code from this PR to #169 which now only covers rto.go

PR #156, opened 2026-07-14
testing.B.Loop() standardization and StackAsync.Debug improvement

soypat:
Basically enables heap allocation debugging when calling Debug methods on StackAsync, which is something I'm interested in adding since enabling `debugheaplog` build tag makes the software very slow on small targets.

### DIFF
--- a/internal/debug.go
+++ b/internal/debug.go
@@ -2,16 +2,16 @@ package internal
 
 import (
 	"log/slog"
+	"os"
 	"runtime"
 	"strconv"
 	"sync"
-	"unsafe"
 )
 
 const (
 	LevelTrace slog.Level = slog.LevelDebug - 2
 
-	usePrintLogAllocs = true
+	usePrintLogAllocs = HeapAllocDebugging
 )
 
 var (
@@ -19,11 +19,11 @@ var (
 	lastAllocs  uint64
 	lastMallocs uint64
 	allocmu     sync.Mutex
-	allocbuf    [256]byte
+	allocbuf    [128]byte
 )
 
-func LogAttrsAndAllocs(l *slog.Logger, level slog.Level, msg string, attrs ...slog.Attr) {
-	logAttrsAndAllocs(l, level, msg, attrs...)
+func LogAttrsAndAllocs(allocmsg string, l *slog.Logger, level slog.Level, msg string, attrs ...slog.Attr) {
+	logAttrsAndAllocs(allocmsg, l, level, msg, attrs...)
 }
 
 func LogAllocs(msg string) {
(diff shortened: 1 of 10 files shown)


MDr164:
Fails to compile for me when I use the debug flag:
```
$ go build -tags=xnetdebug ./x/xnet/
x/xnet/xnet_debug_pcap.go:17:5: undefined: internal
x/xnet/xnet_debug_pcap.go:18:3: undefined: internal
```

In addition: There seems to be an issue currently with the `debugheaplog` buildflag. In the default case e.g. `!debugheaplog` there is a call in StackAsync.Debug and StackAsync.DebugErr to LogAllocs unconditionally which could reduce performance, no? I think there might also be some footguns with the logging, like the [ALLOC] lines going straight to stderr which bypasses slog and LogAllocs runs before LogAttrs so allocations made by the LogAttrs call get attributed to the next call instead. Overall just minor stuff to eventually look into apart from the compile error (I think it's just missing the import)

soypat:
Good catches! Note that LogAllocs performs no I/O if there are no allocations between calls so it should be seen as a lightweight call unless your application is allocating all the time, in which case you may have bigger problems. That said I've taken your comment about stderr regarding println into account and by default LogAllocs will print to stdout unless debugheaplog is enabled, in which case a different println path will be used since that path is known to work exceedingly well. The alternative allocbuf path has had issues in the past if i recall correctly.

internal/debug.go
```diff
@@ -2,50 +2,65 @@ package internal
...
+
 func LogAllocs(msg string) {
 	allocmu.Lock()
 	runtime.ReadMemStats(&memstats)
 	if memstats.TotalAlloc == lastAllocs {
 		allocmu.Unlock()
 		return
 	}
+	inc := int64(memstats.TotalAlloc) - int64(lastAllocs)
+	numAlloc := int64(memstats.TotalAlloc) - int64(lastAllocs)
```
MDr164: I think that's a copy/paste error as it's identical to the line above. This should probably be mallocs - lastAllocs instead of the total?

internal/debug.go
```diff
@@ -2,50 +2,65 @@ package internal
...
-		n += copyValueInt(allocbuf[n:], "inc", int64(memstats.TotalAlloc)-int64(lastAllocs))
-		n += copyValueInt(allocbuf[n:], "n", int64(memstats.Mallocs)-int64(lastMallocs))
+		n += copyValueInt(allocbuf[n:], "inc", inc)
+		n += copyValueInt(allocbuf[n:], "n", numAlloc)
 		n += copyValueUint(allocbuf[n:], "heap", memstats.HeapAlloc)
-		n += copyValueUint(allocbuf[n:], "free", memstats.HeapSys-memstats.HeapInuse)
+		n += copyValueUint(allocbuf[n:], "free", free)
 		n += copyValueUint(allocbuf[n:], "tot", memstats.TotalAlloc)
-		println(unsafe.String(&allocbuf[0], n))
+		allocbuf[n] = '\n'
```
MDr164: If n is 128 this line will panic now I think. Also I don't think the n > len(allocbuf) check further down will ever activate. If I recall correctly the `copy` function will clamp to its destination size and the copyValueXXX calls return 0 on insufficient space. But the bigger issues is the panic and array out-of-bounds access when your message is too large. The newline (and carriage return or line feed for that matter) should be inserted at 127 max aka `len(allocbuf)-1` or so

MDr164 (commented):
Looks better now, found two quick to fix issues.

internal/debug.go
soypat: Great suggestions. applied!

internal/debug.go
soypat: holy cow yes, manual copy paste was super lazy. thanks for the review.

MDr164 (approved):
I think that captures everything now if I haven't missed anything 👍

soypat:
Alright! Merging and will mark this as v0.2.1. Thank you @MDr164 !!!

Issue #161, opened 2026-07-20
CI check to keep block binary blobs from being checked in


soypat:
Checking in binaries can happen to anyone 👀 https://github.com/soypat/lneto/pull/160#issuecomment-5015713449

I imagine this can be easy to catch by iterating over all files and checking that no files start with Linux/Windows/MacOS ELF/AppImage/exe/dmg headers.

MDr164:
There's also a gitignore trick that I usually do which is:
```shell
# Ignore all
*

# Unignore all with extensions
!*.*
``` 

That way files without any extension are gitignored which most of the time are binaries. Not perfect but so far a nice little trick that doesn't need CI additions and catches several accidental binary additions.

PR #162, opened 2026-07-20
@MDr164 trick to ignore no-extension files

soypat:
Solves #161


b0ch3nski (approved):
clever

MDr164:
One caveat right now: The LICENSE file is not named LICENSE.md so any changes to it will now be ignored. Otherwise I think it won't cause any problems right now.

soypat:
Thank you for review!

PR #163, opened 2026-07-21
Add RFC 7323 timestamp negotiation and RTT sampling

MDr164:
Follow up to #155 and split off from #115. This was part of the TCP congestion/loss recovery and is now a standalone PR.

### DIFF
--- a/tcp/conn.go
+++ b/tcp/conn.go
@@ -82,10 +82,16 @@ type ConnConfig struct {
 	// recovery. See [LossRecovery].
 	LossRecovery LossRecovery
 	// Nanotime is the monotonic time source in nanoseconds (the func() int64
-	// convention used across lneto) that drives LossRecovery. It is required when
-	// LossRecovery is set and unused otherwise. The tcp package reads it only to
-	// stamp the loss-recovery hooks; it holds no clock itself.
+	// convention used across lneto). It is the connection's TCP clock: it drives
+	// LossRecovery and generates the RFC 7323 timestamp clock when
+	// EnableTimestamps is set. Required when either LossRecovery or
+	// EnableTimestamps is set, and unused otherwise. The tcp package reads it only
+	// to stamp segments; it holds no clock itself.
 	Nanotime func() int64
+	// EnableTimestamps opts into the RFC 7323 Timestamps option, negotiated on
+	// the handshake and used to measure the round-trip time. It requires Nanotime
+	// (else Configure returns an error). See [Handler.EnableTimestamps].
+	EnableTimestamps bool
 }
 
 // Configure should be called on any newly created connection before usage. See [ConnConfig].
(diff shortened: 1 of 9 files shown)


MDr164:
I assume that RFC7323 is not really that important in total; the first half is in this PR and the second half of that RFC is in part 7 aka the last part of the patch train. I'll leave it up here for now as it's now rebased against the previous changes.

MDr164:
@soypat I closed the other PRs again. I'm now extracting everything I had previously for LossRecovery/CongestionControl into smaller chunks, this one being the first one of those (I can also open all three commits in this branch as inidividual PRs if you prefer). I'll reintroduce the PRs for CUBIC and BBR later but especially BBR seems to be a bit tricky to do, I'm currently trying to figure out a decent way to match it to the new interface. You can start reviewing this one while I try to make the tests pass again after rebase.

soypat (commented):
Let's move the blast radius of this outside the tcp package. Like I said, the Loss interface may be better suited as a `tcp.Driver` interface and have the PreTx method return a self-owned slice of TCP options to be added to the outgoing segment. This way the TCP option logic lies outside of tcp package and I rest knowing users can fiddle with TCP options however they please since it is part of the interface they provide, and not my responsability. I like this path because TCP works well enough for most of lneto's users today and trying to cover everyone's use case in tcp package is going to be a fustercluck if we don't draw the line one what really belongs in there. `tcp.Driver` (today tcp.LossRecovery`) is the abstraction I believe will save us from this quagmire of responsability and complexity. What do you think?

MDr164:
That sounds reasonable but it's not 100% possible as of right now. Especially with me trying to get BBR working I stumbled upon more shortcomings of the existing interface (no changes but additions needed). The same will be true for this PR, I can move a lot into a different package but not all of the hook points and plumbing will work right now. How about this approach then: I do a more thorough clean-up of the current TCP package (and potentially break the API again but as minimal as possible) and move everything out into subpackages that I can already (no additions, just what we already got in). Then I'll analyse what interfaces/hooks are missing as of right now and add them. Afterwards 90%+ of my additions will not be inside the TCP core package anymore. Would that work for you? I can also spend some more time on CI so that all the future changes do not introduce some silent regressions instead.

MDr164:
Want me to commit to creating new interfaces factoring things out of the TCP package or rather not for now? I can also work more on VLAN and IPv6 if you need more time to think about that :sweat_smile: 

For my local usage I eventually need all of the above but I got the basic demo for the upcoming conferences already working with the current state so it's not pressing

MDr164:
Closing this either way as it does not align with the desired final outcome and will make things harder to move in the future.

soypat:
> do a more thorough clean-up of the current TCP package (and potentially break the API again but as minimal as possible) and move everything out into subpackages that I can already ...

Feel free to. I will closely examine changes to API and be in favor of not breaking unless the changes you propose are sound and bring meaningful benefits. I think we can do a v0.3.0 if that is the case. I do urge you to think the change through with the mindset I have been pushing for though. It seems like we keep going in circles. Anyways you seem to have the direction clearer than I do so I'll let you take the next step :)

If you are worried about future breakage with all future features merged:

 **_Don't._** 
 
 It's better to take this step by step, little by little and break stuff slowly. Please do not send a multi-hundred liner PR my way. I much prefer the outline I've given above: Just add the fields you need to Rx/Tx for now, we don't need to think of every case in the future right now. Taking it slow is best in these cases I feel. Again, SLOW IS BETTER. Chill with the future requirements, and just lets go with the flow. Don't think of everything at once, take it slow.

> _The interface will be revealed to the student once they are ready. Never before._

-Confucius or some clanker

PR #165, opened 2026-07-21
TCP congestion part 5: Congestion interface definition

MDr164:
Follow up to #164 and split off from #155

### DIFF
--- a/tcp/definitions.go
+++ b/tcp/definitions.go
@@ -10,6 +10,34 @@ import (
 	"github.com/soypat/lneto"
 )
 
+// CongestionControl decides how the TCP sender reacts to congestion.
+//
+// The implementation may inspect the current handler state and the most recent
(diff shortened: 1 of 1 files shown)


MDr164:
@soypat I assume the loss interface fully supersedes this one then?

soypat:
Short answer: Yes. 

Long answer: Whether we arrive at Loss or a more complete interface like `tcp.Driver` I want it to be a single interface that adds timing and more complex logic for the reasons I've outlined: tcp already works really well for ! Lets move the customization users may want to timing, retransmission and options to an optional dependency injection interface, if it makes sense it'll be a single interface. So far I want to push for a single interface design.

MDr164:
Then I'll close this one and push the decision out aka use the LossRecovery interface for part 6 for the patch train. Thanks for the feedback! :)

PR #168, opened 2026-07-22
add loss.go

soypat:
@MDr164 This is an example of how one could structure the new loss API change. Things I like about this way forward:

- One package is touched by this change. Great opportunity to write tests for this single change and communicate intention with it. Future changes that go back on intention will break these tests. Very nice for several engineering reasons
- There is no unused API. I've purposefully omitted the unused fields on Directives, though I have left them commented in to note future plans.
- RetransmitAll instead of Retransmit is clearer in intention.
- Nit: Field ordering in structs to favor compactness
- A Method for h.loss != nil that communicates intention
- Avoid duplicated comments i.e: `Persist configuration across reopen:`, notice how I've grouped the fields that share this comment semantics together

Note: The tests are fully AI generated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # Ignore files with no extensions trick:
 # First ignore all.
 *
+# Unignore directories since they usually have no extensions
+!*/
 # Unignore all with extensions.
 !*.*
 !LICENSE
--- a/tcp/conn.go
+++ b/tcp/conn.go
@@ -76,12 +76,26 @@ type ConnConfig struct {
 	// Logger sets the [Conn] logger.
 	// Lower level logging available at [Handler.SetLoggers] via [Conn.InternalHandler].
 	Logger *slog.Logger
+	// LossRecovery is the optional packet-loss recovery algorithm (RTO,
+	// congestion control, ...) for the connection. If set, Nanotime must also be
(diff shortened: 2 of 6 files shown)


MDr164:
Ok I see. Well it does change a couple of my PRs but not all of them actually. E.g. part 4 stays, for part 3 I need to rethink how I do it without RTTObserver, I suppose part 5 is then completely dropped (that was the previous interface we defined way back)? Part 6 needs rework on top of this interface then and 7 stays for the most part as well. And part 2, the one that started this discussion is then largely obsolete by this PR here I guess.

soypat:
Alright, let's take it one step at a time. Keep changes organized and easy to follow and I think we'll be merging them in no time. I can't speak for the parts after 2, but if it is anything like part 2 then I'd encourage you to put in special effort to really finding the base unit of functionality you want to add and polishing it well. The TCP package is already quite complex and I will not compromise on merging things I struggle to grasp when they can be broken down into smaller work packages. I'll go ahead and merge this since it's pretty much your changes

MDr164:
Sounds good! I'll rework the parts with this now in today so expect some updates in a couple hours. Please don't bother looking all of the draft PRs. When I leave something in draft it really means do not bother looking at it yet :)

Should I close the part 2 PR then?

soypat:
Alright! Affirmative on all parts

PR #170, opened 2026-07-25
DHCPv4+Debug

soypat:
add RawData method and improve Logging semantics with configured logger instead of default


soypat:
Small enough change. want to test out heap allocs in field so merging

PR #171, opened 2026-07-25
add http/httphi

soypat:
New httphi package. Hand designed with minimal AI intervention to be extremely memory efficient. Still needs testing and adding a few more convenience methods.


MDr164:
Oh wow! That's a huge PR but very impressive work. I'll give it a try on Tamago and see how it performs :)

soypat:
That would be hugely appreciated! 
Currently working on benchmarks which I think I'll submit as a separate PR on top of this one

MDr164:
I found a couple issues that occur when not following the happy path so it needs a bit of hardening of the errors paths. I'll create a more detailed review once I understand their actual origin (there seems to be a goroutine leak for example) and if I spot anything else. :+1: 

But I do like the general design so far

soypat:
Found a bunch of issues while writing the benchmarks. Pushing fixes and improvements- reason CI is failing is because I'm currently working alongside many benchmark changes I don't want to push yet so the repo view is not consistent

MDr164:
Alright, let me know when you think this is good to go and review. I found the biggest issues are in the Configure/Reconfigure flow as it creates inconsistent state when called in the middle of things. I'll continue with the TCP rework and making it as non-invasive as I can.

soypat:
This is good to review! No more issues cropped up during benchmarking. Got cool results

heap-per-request
latency

examples/berkeley-listener/berkeley_server.go
```diff
@@ -370,7 +370,7 @@ func mockClient(stack *xnet.StackAsync, port uint16, subnet netip.Prefix) {
 
 	var hdr httpraw.Header
 	hdr.SetMethod("GET")
-	hdr.SetRequestURI("/")
+	hdr.SetRequestTarget("/")
```
MDr164, line 373: I do like the naming better but technically it's breaking the API. Do you want to tag a new version or not? I don't really care as I can quickly adapt my code but might be good to know in general in case that I need to change API in e.g. TCP

http/httphi/router.go
```diff
@@ -0,0 +1,362 @@
...
+		return nil
+	}
+	return lneto.ErrPacketDrop
+}
+
+func (r *Router) goroWorker(gen uint32, queue chan job, backoff lneto.BackoffStrategy, mux Mux) {
+	for job := range queue {
+		exch := job.exch
+		if gen != r.gen.Load() {
+			return
```
MDr164: Reconfiguration can permanently leak queued exchanges and connections it seems. `TeardownGoroutines` increments the router generation and closes the worker queue. A worker that subsequently dequeues an old job sees a generation mismatch and returns immediately in this line. It does not call `r.freeExch`, release the exchange, or close the connection. A `sync.WaitGroup` or generation-specific completion object would be substantially safer than polling the `used` flag imo.

http/httphi/router.go
```diff
@@ -0,0 +1,362 @@
...
+	reqBuf          int
+	respBuf         int
+	reqNumHeaderCap int
+	normalizeKeys   bool
+	pendingConns    chan job
+	mux             Mux
+
+	globbuf  []byte
+	exchs    []Exchange
+	freeList *Exchange
```
MDr164, line 62: This field is maintained during different sets of `Configure` in the dynamic branch and only cleared in the fixed-worker branch of code. There is also an issue if you image following scenario:
1. A dynamically allocated exchange is serving.
2. `Configure` installs a new configuration.
3. The old handler finishes.
4. `freeExch` inserts the old exchange into the current router free list.
5. A request under the new configuration reuses that old exchange and its old buffer sizes.

`freeExch` indiscriminately inserts exchanges into the shared list, while `getExchLocked` uses the list before inspecting the current pool.

MDr164 (commented):
There are a couple more weird cases I found while hammering the code a bit but I'll first have to confirm them

examples/berkeley-listener/berkeley_server.go
soypat: Yeah this'll be v0.3.0. An upgrade from 1999 naming convention :)

soypat:
@MDr164 Ready for another review!

MDr164:
Looks better but I did find a couple more items, one panic and a couple smaller items all on the error/non-default path again (Content-Length is case sensitive and malformed values in it also make parsing misbehave). I'll try to create another more detailed review inbetween working on the TCP code experiment.

MDr164:
On a sidenote: Would it make sense to also allow your new httpbench tool to use lneto httptap as a comparison point? Trying to find a good way to fully test the performance of the HTTP stack e2e using also the lneto TCP+HTTPhi combo.

soypat:
Absolutely! Feel free to submit a PR to httpbench to replace the underlying TCPListener with lneto stack

deadprogram:
Any way to add an `examples_test.go` to the `httphi`package  that show some introduction on how it shouold be used? Maybe a `README.md` would also help since it will end up in the Go pkg docs as well.

soypat:
Done! @deadprogram

http/httphi/router.go
```diff
@@ -0,0 +1,378 @@
...
+	if workerMode && cfg.MaxAwaitingConns == 0 ||
+		cfg.Mux == nil ||
+		cfg.RequestNumHeaderKVCap <= 0 ||
+		!workerMode && cfg.FixedNumGoroutines != -1 {
+		return lneto.ErrInvalidConfig
+	} else if cfg.Backoff == nil {
+		return lneto.ErrMissingHALConfig
+	}
+	return nil
+}
```
MDr164, line 141: I think it's missing checks for:
- Negative MaxAwaitingConns.
- Non-positive request and response buffer sizes.
- Minimum response space for the terminal CRLF.
- Addition and multiplication overflow.
- Buffers incompatible with the uint16 offsets used by Exchange.
- Excessively large worker counts or total allocations.

Not sure if I missed some potential oddities to put into a config.

http/httphi/exchange.go
```diff
@@ -0,0 +1,638 @@
...
+func (exch *Exchange) RequestParseCookie(dst *httpraw.Cookie, key string) error {
+	value := exch.RequestHeader(key)
+	return dst.ParseBytes(value)
+}
+
+// RequestContentType returns the request's Content-Type field value as it
+// appears on the wire, parameters included, nil if absent. Test it with
+// [httpraw.MediaTypeIs] and pick parameters out with [httpraw.ContentParam].
+func (exch *Exchange) RequestContentType() []byte {
+	return exch.RequestHeader("Content-Type")
```
MDr164: What if this is CONTENT-TYPE instead e.g. a weird server sends this in all-caps or lowercase? The same for other fields like Content-Type and so on. Right now all of this is case sensitive and while it works for the most part I've seen quirky servers out there replying different than expected.

http/httphi/exchange.go
```diff
@@ -0,0 +1,638 @@
...
+		return errNotFormEncoded
+	} else if exch.RequestHeader("Transfer-Encoding") != nil {
+		// Chunked bodies are framed, so reading Content-Length bytes off the
+		// wire would parse chunk sizes as form data. httpraw does not decode them.
+		return errUnsupportedTransferCoding
+	}
+	length, err := exch.RequestContentLength()
+	if err != nil {
+		dst.Reset(buf[:0])
+		return dst.Parse() // No length is no body, RFC 9112 6.3.
```
MDr164: You only check if the err is not nil, but if the error returned is not an absent content-length but rather a malformed one then it's treated as if the body is empty. Is that the right behavior or should that be handled differently? E.g. things like:
```
Content-Length: garbage
Content-Length: -1
Content-Length: 1, 2
Content-Length: 999999999999999999999999
```
all trigger this branch

http/httphi/exchange.go
```diff
@@ -0,0 +1,638 @@
...
+//	}
+func (exch *Exchange) HijackRaw(dstBody []byte) (conn, []byte, error) {
+	data, err := exch.remainingSurplusBody()
+	if err != nil {
+		return nil, nil, err
+	}
+	exch.hijacked = true
+	dstBody = append(dstBody, data...)
+	return exch.rw, dstBody, nil
+}
```
MDr164, line 103: This only sets `hijacked` to true but should probably do more because: After the handler returns, `Handle` still unconditionally calls `FlushHeader`. If the handler did not already write a response, that emits a default `HTTP/1.1 200 OK` onto the hijacked stream. `ExchangeRW.IsValid` still ignores `hijacked`, even though its documentation states that a hijack invalidates the handle.

MDr164 (commented):
You dropped some of the backoff code after the last review I did and I think `RequestParseForm` can now busy-spin indefinitely, no? Also found the items I made inline comments for. I think there is also a problem that httphi accepts malformed protocols like HTTP/6.9 and only rejects empty ones while httpraw was only accepting HTTP/1.1 explicitly iirc. I think that's all I have for now, still busy wrapping my head around the TCP experiments 😅

dgryski:
I would love to see some fuzz tests for all the parsing routes, as well scaffolding for fuzzing full requests from a server.

http/httphi/exchange.go
soypat: Good catch with the FlushHeader! I've removed the IsValid documentation referring to hijacked, which seemed to open a can of worms if we wanted to enforce that.

http/httphi/exchange.go
soypat: Good catch, fixed some clankerlicious code right there. Still need to do something with the Form.Parse returning an error when they are completely unable to

http/httphi/exchange.go
soypat: Good catch, added GetFold and added a bool to signal the content length field was present.

http/httphi/router.go
soypat: Added checks! Thanks for the heads up!

soypat:
@dgryski  Even the fuzz testing is quite lightweight. Steady 70MB at 200kreq/sec per thread  :P

end to end fuzzing belongs somewhere else though. I think adding a fuzz subcommand to [httpbench](https://github.com/soypat/httpbench) would be a great addition.
```
fuzz: elapsed: 8m0s, execs: 89599056 (191540/sec), new interesting: 40 (total: 158)
fuzz: elapsed: 8m3s, execs: 90159691 (186906/sec), new interesting: 40 (total: 158)
fuzz: elapsed: 8m6s, execs: 90722517 (187548/sec), new interesting: 40 (total: 158)
fuzz: elapsed: 8m9s, execs: 91255188 (177580/sec), new interesting: 40 (total: 158)
fuzz: elapsed: 8m12s, execs: 91800605 (181814/sec), new interesting: 40 (total: 158)
fuzz: elapsed: 8m15s, execs: 92358007 (185823/sec), new interesting: 40 (total: 158)
fuzz: elapsed: 8m18s, execs: 92925664 (189224/sec), new interesting: 40 (total: 158)
```
image

soypat:
>You dropped some of the backoff code after the last review I did and I think RequestParseForm can now busy-spin indefinitely, no? 

Yes! io.Reader implementation should block! Probably should document that in Exchange.Acquire and Router.Handle...

> I think there is also a problem that httphi accepts malformed protocols like HTTP/6.9 and only rejects empty ones while httpraw was only accepting HTTP/1.1 explicitly iirc. 

Hmmm... should this be an error? httpraw is used by httphi so they should accept same protocols- i think...

MDr164:
> end to end fuzzing belongs somewhere else though. I think adding a fuzz subcommand to [httpbench](https://github.com/soypat/httpbench?rgh-link-date=2026-07-29T00%3A44%3A32Z) would be a great addition.

I would argue that ltesto is a better place than httpbench. Fuzzing is not what I would particularly associate with a benchmark tool.

> Hmmm... should this be an error?

I'm actually not sure, it falls back to HTTP/1.1 rn iirc which might be better than an error. Just wanted to point out that there seems to be a slight UX different between httphi and httpraw as of right now.

I'll take another look but if the previous issues I found are fixed it should be good to go imo

soypat:
After doing some reading I've opted for early rejection of unsupported HTTP protocols. Think there is sense in that since it'd result in garbage being read on user's side.

> I would argue that ltesto is a better place than httpbench. Fuzzing is not what I would particularly associate with a benchmark tool.

I'd like to have fuzzing engine in ltesto that could be reused- maybe we can extract the fuzzing engine in httphi to ltesto to eventually let others import it(?).

The reason I mention httpbench is because @dgryski mentioned fuzzing end-2-end with actual network connections. That is something I feel does not belong here in lneto for now. httpbench is a good place for it since it could also test other HTTP implementations like fasthttp. In any case we can rename httpbench to httptest/httpsuite/httptorture if the name is not fitting.

http/httphi/mux.go
```diff
@@ -0,0 +1,217 @@
...
+// replies 404 when it returns nil.
+type Mux interface {
+	// LookupHandler matches the requestPath and method to a handler and returns it and the
+	// pattern it matched.
+	LookupHandler(get Method, requestPath string) (matchedPattern string, handler HandlerFunc)
+}
+
+// MuxSlice is a [Mux] backed by a slice of registered endpoints, matched by
+// exact path. Lookup is linear in the number of registrations.
+type MuxSlice struct {
```
deadprogram, line 109: `MuxHandlers` maybe? `MuxSlice` just seems like an odd term, I do not know why.

soypat: It's an implementation that uses linear lookup, that's what Slice implies. Could call it MuxLinear, but I feel that is odder. I'd rather point to the underlying data structure for the Mux implementation

soypat:
Added very neat PathValue parsing to match Go's ServeMux implementation and lots of documentation and rephrased some of the method names regarding Request queries. I think we are ready to merge this ASAP. Also fixed a bad test in rawsock package.

MDr164:
I'll take a quick look but if you wanna merge it already go ahead and we can follow up with any fixes afterwards like with netdev

http/httphi/router.go
```diff
@@ -0,0 +1,406 @@
...
+
+func (cfg RouterConfig) workerMode() bool {
+	return cfg.FixedNumGoroutines > 0
+}
+
+// TeardownGoroutines stops the router's fixed goroutines once they finish the
+// exchanges they are serving. [Router.Configure] calls it before installing a
+// new generation. A torn down router refuses connections with a non-nil error
+// until it is configured again.
+func (r *Router) TeardownGoroutines() {
```
deadprogram: Couldn't this being something more friendly like `Stop()` maybe?

soypat:
Alright, I'm going out to drink some well deserved coffee! Feel free to review, will probably look at merging within ~2 hours or so unless new unaddressed items appear! We can always fix stuff in future :) 

(also good opportunity to leave the fuzzer running)

MDr164:
Sounds good, from a quick glance everything in the code I already looked at is resolved. I'll take a loot at the new additions like the router and kv stuff that's new now.

EDIT: There might be some smaller leftovers like content-type and transfer-encoding still being case sensitive and it's midnight over here so I can't really give it a super thorough review round. I suppose we can push some follow up fixes if needed.

soypat:
Fuzzer's happy and so am I: `fuzz: elapsed: 2h11m3s, execs: 672937654 (86491/sec), new interesting: 60 (total: 176)`

Thanks to everyone who helped review! @MDr164 @dgryski @deadprogram  many many thanks! 

Will merge in this state so we can start thinking about where to use httphi!

PR #172, opened 2026-07-28
tlsraw: a low level TLS package focused on readability and security

soypat:
`tlsraw` is a mostly human written package with heavy focus on readability and security. Performance is secondary... although still took care of those pesky heap allocations. I want this to run without GC pressure at runtime


MDr164:
Ah nice, I already saw that branch pop up on the repo. I think that revives the discussion of whether or not we want crypto stuff in here or just minimal shims to plug in some crypto from stdlib.

soypat:
TODO: was using mbedtls and wolfssl as reference but now have learned of https://gitlab.com/devnw/zig/tls/ and thinking about using it as reference. Looks to have very nice testing methodology. Cheers @benjivesterby

soypat:
@MDr164 Pinging you since you had reviewed a while ago. I've reduced the scope of the PR to only implement parsing, decoding, encoding of records. This is as low level TLS as you can get without implementing server/client state and any of the x509 stuff. Based on the code claude generated for a TLS server, that'll probably be included in the `x` package as super duper at-your-own-risk later since I found it much harder to follow than this... of course maybe I can find a way of simplifying it... who knows.

Edit: ~The binary size is a big issue- MWE should absolutely not be bringing in crypto libraries :0~ Solved! crypto has some `init()` function calls which include tens of kB... for fips compliance maybe?

Edit2: will probably start work on a branch on top of this PR soon on a new package that pulls from `tlsraw`, `tlsauto` maybe

MDr164:
I'll give it a thorough review after returning from OSFC travel on Monday. So far it looks generally pretty good with a couple of cases I need to double check once I got more time to do so. With the interface additions I might be able to even drop stdlib/crypto usage from NTS.

soypat:
OK, the interfaces look much improved now. should support diffie-hellman and ml-kem key exchanges

MDr164:
I'll give it a more detailed review then if you think the current state is about done, otherwise I'll wait until it settles and look at the DNS/CNAME stuff again in the meantime

soypat:
DNS/Cname higher priority. As for this branch I'm in no hurry- and I'll stop touching the API. Quite happy with how it turned out

crypto/internal/lcrypto/lcrypto.go
```diff
@@ -0,0 +1,172 @@
...
+type KeyExchange interface {
+	// ID returns the RFC 8446 B.3.1.4 NamedGroup wire value.
+	ID() uint16
+	// NewExchanger returns an unkeyed Exchanger. Callers build one per
+	// connection and call ClientGenerateRekey or ServerSharedRekey to reuse.
+	NewExchanger() Exchanger
+	// SharedLen returns the length of the client key_share, the server
+	// key_share (a ciphertext for KEM groups) and the shared secret.
+	SharedLen() (clientShare, serverShare, shared int)
+}
```
MDr164: In my opinion those interfaces being `internal` is going to bite us here if these are meant as provider interfaces. `Suite.NewAEAD() AEADCipher` and `KeyExchange.NewExchanger() Exchanger` require downstream implementations to name return types from crypto/internal/lcrypto, which Go won't let packages outside the lneto/crypto tree import. So external crypto/HSM/etc implementations can't actually implement these factory interfaces. Not a problem if that is intended and no external entity is meant to actually use those but I'm not the biggest fan tbh. Could we either move the provider-facing interfaces somewhere exported or avoid internal types in the factory method signatures? 👀

crypto/tlsraw/keyschedule.go
```diff
@@ -0,0 +1,181 @@
...
+// Master advances to the master secret and writes the application traffic secrets.
+// Call after adding the server Finished.
+func (ks *KeySchedule) Master(client, server *[32]byte) {
+	ks.advance(zeroSecret[:])
+	ks.trafficSecrets(client, server, "c ap traffic", "s ap traffic")
+}
+
+// Keys writes the record protection key and IV of a traffic secret, RFC 8446 7.3.
+//   - AES128: len(key)==16
+//   - AES256/ChaCha20: len(key)==32
```
MDr164, line 91: I think the AES256 note is misleading here. `KeySchedule` is still hard-wired to SHA-256 and 32 byte secrets and Reset explicitly rejects anything else, so e.g. `TLS_AES_256_GCM_SHA384` can't actually use this even though Keys now accepts a 32 byte key. ChaCha20/SHA256 is fine. I'd either drop AES256 from the comment for now or generalize the schedule. Or I might have mixed up things myself, it's a 3k LoC PR so I might overlook things myself.

crypto/tlsraw/record.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	hc.seq++
+	return nil
+}
+
+// HasKeys reports whether SetAEAD installed keys since the last Zeroize.
+func (hc *HalfConn) HasKeys() bool { return hc.aead != nil }
+
+// Zeroize forgets the keys. SetAEAD must be called before reuse.
+func (hc *HalfConn) Zeroize() { *hc = HalfConn{} }
```
MDr164, line 111: We should actually call `hc.aead.Zeroize()` here before clearing the struct. The new interface specifically adds `Zeroize` so an implementation can wipe its key schedule/state, but `HalfConn.Zeroize` currently just drops the interface reference. If the caller keeps another reference or the AEAD uses static storage the key is still alive in memory. On that note, there should probably be a note that the crypto stuff in here is not externally audited. While I'm not the worst at spotting holes I'm also not the best nor accredited to do so :D

crypto/tlsraw/record.go
```diff
@@ -0,0 +1,104 @@
...
+	// Content type is the last non-zero byte; zeros after it are padding.
+	i := len(plain) - 1
+	for i >= 0 && plain[i] == 0 {
+		i--
+	}
+	if i < 0 {
+		return nil, 0, lneto.ErrInvalidField
+	}
+	return plain[:i], ContentType(plain[i]), nil
+}
```
MDr164, line 90: I think the zero-value path still panics here. Both `Seal` and `Open` call `hc.aead.Overhead()` before checking whether keys were installed, so using a fresh or zeroized `HalfConn` gives you a nil interface panic instead of an error. Since `HasKeys()` already exists I'd just reject that case with `ErrInvalidConfig` first which _should_ suffice (famous last words).

crypto/tlsraw/hello.go
```diff
@@ -0,0 +1,493 @@
...
+		if sentByServer {
+			// A server acknowledges the name with empty extension_data.
+			if len(data) != 0 {
+				err = lneto.ErrInvalidLengthField
+			}
+		} else {
+			err = validateServerNames(data)
+		}
+	case ExtALPN:
+		err = validateALPN(data)
```
MDr164: Either I'm paranoid or the server ALPN validation still looks too permissive to me. We pass `sentByServer` into the extension validation but then validate ALPN exactly like the client list, so a server response containing multiple protocol names is accepted. The server is only allowed to select one afaik. Can we make the server case require exactly one entry or am I wrong?

crypto/internal/lcrypto/lcrypto.go
```diff
@@ -0,0 +1,172 @@
...
+	//  - scheme is the RFC 8446 B.3.1.3 SignatureScheme of the peer's CertificateVerify. The caller has
+	//    checked it is one it offered; the implementation must check it suits the leaf's key type.
+	//  - serverName is the name the leaf must match, the SNI sent by the client. Empty when peerIsServer=false.
+	//  - msg is the complete, unhashed RFC 8446 4.4.3 content: [64B pfx, context string, 0x00,
+	//    handshake transcript hash], built by the caller. Implementations verify msg as given.
+	//  - sig is the signature from the peer's CertificateVerify, remote and potentially adversarial.
+	//
+	// chain, serverName, msg and sig are only valid for the duration of the call and must not be
+	// retained or modified: they point into the caller's buffers, which are reused by the next connection.
+	VerifyPeer(chainView CertChain, peerIsServer bool, scheme uint16, serverName, msg, sig []byte) error
```
MDr164, line 124: I think we should avoid defining `serverName` here as specifically the SNI value. SNI can legitimately be absent while we still have an expected peer identity to verify, most obviously when connecting to an IP address directly. If empty SNI turns into an empty name check here, that can accidentally turn off endpoint verification. Can we make this the configured expected peer name/IP instead and keep it separate from whatever was actually sent as SNI? Haven't checked how other stacks do it but in the BMC world we often don't have host/server names.

MDr164 (commented):
Talking about large PRs I think I chewed threw this one now. High chance I missed stuff. Though most of it looks pretty sane.

crypto/internal/lcrypto/lcrypto.go
soypat: snap, did not see that one coming. Will think of something to fix this. You are absolutely right.

soypat:
Excellent comments @MDr164. Will likely sit down by wednesday to take the time to fix these carefully

crypto/internal/lcrypto/lcrypto.go
soypat: OK... this stuff is new to me- is the text below an accurate diagnosis?

<details>

I read the interface, its callers, and the stdlib equivalents. The concern is correct, and it's sharper than "paranoid" — stdlib treats these as two separate values by construction, and the specific case named (dialing an IP) is the one where they provably cannot be the same value.

## The two things being conflated

[lcrypto.go:117](crypto/internal/lcrypto/lcrypto.go#L117) defines the parameter by its wire provenance: *"serverName is the name the leaf must match, the SNI sent by the client."* That sentence fuses two independent concepts:

- **SNI** — a routing hint. Client→server, cleartext, on the wire, whose only job is letting a multi-tenant server pick which certificate to present. RFC 6066 §3 constrains it: no literal IPv4/IPv6 addresses, no trailing-dot FQDN.
- **Expected peer identity** — local policy. The identity the application intended to reach, and the input to name matching during verification.

They coincide for the browser case (dial `example.com` → SNI `example.com` → match leaf against `example.com`), which is why the conflation reads as harmless. They diverge everywhere else.

## The failure mechanism

`VerifyPeer` is documented as the sole authority: *"A nil return is the only way a peer becomes authenticated"* ([lcrypto.go:109](crypto/internal/lcrypto/lcrypto.go#L109)), and `Verifier` *"subsumes trust anchors, hostname matching and expiry so the policy cannot be configured apart"* ([lcrypto.go:103-105](crypto/internal/lcrypto/lcrypto.go#L103-L105)). So every identity signal an implementation has must arrive through these parameters.

An implementer reads "the SNI sent by the client", receives a zero-length slice, and correctly concludes: no SNI was sent, so there is no name to check. The idiomatic Go realization of that is `x509.VerifyOptions{DNSName: ""}` — and `crypto/x509/verify.go:589` guards the hostname check behind `if len(opts.DNSName) > 0`, so an empty name skips `VerifyHostname` **entirely**. What remains is chain-to-a-trusted-root, expiry, and key usage. That accepts any valid certificate for any domain under any trusted CA. Against an on-path attacker — the exact adversary certificate authentication exists to stop — that is a total break of endpoint authentication, and it is silent: the handshake completes, `VerifyPeer` returns nil, the connection reports as authenticated.

So the risk isn't "verification gets weaker." It's that an absent *routing hint* silently disables *authentication*, through a path that looks correct to whoever implements the interface.

## Why the BMC case makes this the default, not an edge case

RFC 6066 §3 forbids IP literals in `HostName`. A conformant client dialing `192.168.1.10` therefore sends **no** `server_name` extension at all — not an empty one, none. There is still an expected identity: the leaf must carry a matching `iPAddress` SAN.

In the BMC/IPMI/Redfish/embedded-management world your coworker is describing, dial-by-IP is the normal case, not the exception. Under the current contract, that population hits the empty-name path on *every* connection. The deployment lneto is aimed at is the deployment where the downgrade is always on.

There's a second, subtler point in that case: the needed identity isn't a name at all. `crypto/x509/verify.go:942-950` shows IP identities match only against `c.IPAddresses` (RFC 6125 App. B.2) — a different matching rule over a different DER SAN type than `dNSName`. A parameter defined as "the SNI" can never carry it, because SNI structurally cannot. This is an expressiveness gap, not only a naming one.

## Why the interface cannot fail closed as specified

Empty `serverName` is currently ambiguous across at least three states:

1. No expected identity configured — verification deliberately off (first-boot provisioning, TOFU pinning).
2. An expected identity exists but is unrepresentable in SNI (IP literal).
3. No name is applicable at all — the doc explicitly mandates this for mutual TLS: *"Empty when peerIsServer=false."*

An implementation receiving empty bytes has no way to distinguish these, and no safe universal rule exists. Reject-on-empty breaks both the IP case and the mTLS case the doc requires. Accept-on-empty is the silent downgrade. The contract makes case 3 mandatory, which forecloses the only fail-closed reading. That's the structural defect: the ambiguity is unresolvable on the receiving side, so no conforming implementation can be written that is safe in all three.

## What stdlib does (verified in the local tree)

`crypto/tls` keeps them as two values with a one-way lossy map, and makes the dangerous state unreachable:

- `handshake_client.go:45-46` — empty `ServerName` with `InsecureSkipVerify` false is a **hard error before any bytes go on the wire**. Emptiness is never reached by accident; disabling verification requires naming a field that says what it is.
- `handshake_client.go:77` — the wire value is `hostnameInSNI(config.ServerName)`, an explicitly lossy *derivation*: IP literals become `""`, brackets and zone IDs are stripped, trailing dots removed (`handshake_client.go:1295-1310`). Config is the source; SNI is the projection.
- `handshake_client.go:1151` — verification uses `DNSName: c.config.ServerName`, the **config** value. The wire value never reaches the verifier.

So the separation your coworker is asking for is exactly the invariant stdlib enforces, and it enforces it at config time rather than trusting the verifier to notice.

## Two smaller observations

**Provenance labeling.** Every other input on this interface is annotated with its trust direction — *"chainView is remote and potentially adversarial"*, *"offered is remote"*, and on `AEADCipher.Rekey`, *"key is local... never peer data"*. `serverName` is the one identity-bearing input with no local/remote label, and the one whose stated origin is the wire. Today `peerIsServer=false` forcing empty closes the case where SNI would genuinely be peer-authored, but the framing invites a later implementer to feed received SNI into an identity check — peer-controlled input steering the peer's own authorization.

**The lifetime clause encodes the wire assumption.** [lcrypto.go:122-123](crypto/internal/lcrypto/lcrypto.go#L122-L123) puts `serverName` in the same bucket as `chainView`/`msg`/`sig`: *"point into the caller's buffers, which are reused by the next connection."* That's the right constraint for a view into a received record, and stricter than a configured value would need — corroborating that the field was designed as a slice into the hello buffer rather than as policy.

## Timing

`VerifyPeer` has no implementations and no call sites anywhere in the repo, and the client handshake is still a TODO ([conn.go:224](crypto/tlsauto/conn.go#L224)); `Config` carries no `Verifier` or expected-name field yet. The contract is the only artifact that exists. That cuts both ways: nothing is broken right now, but `Verifier` is the third-party extension point and the sole authentication authority, so once implementations exist out of tree the assumption is baked into each of them where it can't be corrected centrally.

Per your instruction I've stopped at the analysis and haven't proposed a shape for the fix.

</details>

MDr164: It's a bit inflated and LLM-fluff riddled as usual but generally correct, yes. Right now two things are coupled into one with that API you're proposing and as the trusty datacenter-parrot said what stdlib uses are things not bundled into each other (SNI and serverName). I never looked at how stdlib did it under the hood but the findings all look correct.

PR #173, opened 2026-07-30
Further improvements to httphi API

soypat:
SUMMARY
This branch changes the httphi and httpraw packages. It adds the path
wildcards to the mux. It adds the form reader and the multipart reader
to the exchange. It also gives new names to the HTTP/1 types and to the
examples.

CHANGES TO httphi
- The mux matches the path wildcards. A pattern can contain "{name}",
  "{name...}", and "{$}".
- The mux selects the most specific pattern. It does not select the
  first pattern.
- A fix corrects the path match. Before the fix, the mux did not match
  some paths correctly.
- Mux.MaxPathValues tells the router how many path values each exchange
  must hold. A handler reads one value with Exchange.PathValue.
- Exchange.Respond and Exchange.RespondString write a full response in
  one call.
- Exchange.RequestParseForm reads the pairs from the body, from the
  query, or from both.
- Exchange.ReadMultiparts reads a multipart body. The caller gives the
  buffer and a sink for each part.
- The router removes the MaxAwaitingConns field. The queue length is now
  equal to FixedNumGoroutines. A connection gets an exchange before it
  enters the queue. Therefore the goroutine count sets the limit. The
  queue length does not set the limit.

CHANGES TO httpraw
- The type Header changes its name to HeaderV1. The files header.go and
  parse.go change their names to headerv1.go and parsev1.go. The new
  names show the HTTP version.
- Form gets the methods ReadFromBytes, ReadLimited, and BufferUsed.
  These methods let the caller fill the buffer from more than one
  source.

CHANGES TO THE EXAMPLES
- The examples get new names. Each name shows the package that the
  example uses.
- The example httphi-over-ethernet uses httphi instead of httpraw.
- The example httphi-fuzz is new. It is a server for the ffuf fuzzer.
  The script fuzz.sh runs ten test cases against the server.


soypat:
@MDr164 A little heads up, more breaking changes in API signatures. I think one that may hit hard is httpraw.Header->httpraw.HeaderV1. This will begin making sense after we begin adding V2..V3 HTTP support which is incompatible in how it is handled at the parse level, so good to split out by version.

MDr164:
I did not spot anything critical so I had the clanker take a look:

<details>
<summary>LLM review</summary>

<h3>1. A methodless route can shadow a method-specific route</h3>

`LookupHandler` ranks candidates using only `endpoint.spec`, which is calculated from the path. Once a matching route is found, an equal-scoring later route is skipped. A bare route and a GET route using the same path have identical scores.

This therefore depends on registration order:

```go
mux.Handle("/status", anyHandler)
mux.Handle("GET /status", getHandler)
```

A GET /status request selects `anyHandler`, because the later method-specific registration cannot beat the equal path score. Reversing registration order changes the result.

Fix: include method specificity in candidate ranking. For a request matching both, an exact method should beat MethUndefined. A simple ordering is:

1. Higher path specificity.
2. Exact method over methodless.
3. Registration order only as the final tie-breaker.

Add tests with both registration orders.

<h3>2. Invalid multi-wildcard patterns can silently ignore their suffix</h3>

`MuxSlice.Handle` validates method casing, a leading slash, and exact duplicates, but does not validate wildcard grammar. In `setPathValues`, a multi wildcard returns a successful match immediately without checking whether more pattern segments follow.

For example:

```go
mux.Handle("/a/{rest...}/tail", handler)
```

can match `/a/value`, because `{rest...}` terminates matching and `/tail` is ignored.

Registration should reject at least:

- `{name...}` anywhere except the final segment.
- `{$}` anywhere except the final segment.
- Empty single-wildcard names such as `{}`.
- Duplicate wildcard names in one pattern.
- Malformed brace syntax.

This fits the existing policy of panicking at registration for permanently unusable routes.

<h3>3. An invalid new mux shuts down a valid running router</h3>

`Router.Configure` currently:

1. Locks the router.
2. Calls shutdownLocked.
3. Copies several new configuration fields into the router.
4. Calls cfg.Mux.MaxPathValues().
5. Returns an error when the mux has no registered routes.

As a result, this can destroy a working configuration:

```go
router.Configure(validConfig)
err := router.Configure(configWithEmptyMux)
// err != nil, but the old router has already been shut down.
```

Fix: calculate and validate `maxPathValues` before locking or mutating the router:

```go
maxPathValues := cfg.Mux.MaxPathValues()
if maxPathValues < 0 {
    return lneto.ErrInvalidConfig
}
```

All allocation-overflow checks involving `maxPathValues` should also happen during this preflight phase. A failed `Configure` should leave the previous configuration operational.

<h3>4. Form-body length handling can accept truncation or request huge allocations</h3>

Issue 1: Premature EOF is accepted

`readBodyForm` breaks when it receives `io.EOF` before the declared Content-Length, then returns success and parses the partial body.

A request declaring 100 bytes but sending 10 is therefore interpreted as a valid 10-byte form. This should return `io.ErrUnexpectedEOF`.

Issue 2: Large Content-Length can trigger excessive allocation

The loop converts the parsed int64 length to int and passes the whole remaining length to `Form.ReadLimited`. `kvBuffer.ReadLimited` calls `slices.Grow(kvb.buf, limit)` before limiting the actual read to the 65,535-byte offset range.

Consequences:

- A very large `Content-Length` can request a correspondingly large allocation even though `kvBuffer` cannot address more than 65,535 bytes.
- On 32-bit targets, `int(length)` can overflow.
- The default-growing form path is exposed to a simple memory-exhaustion request.

The code should reject before conversion or growth:

```go
if length > int64(maxFormBuffer-dst.BufferUsed()) {
    return httpraw.ErrBufferExhausted
}
```

The maximum needs to be publicly expressible or enforced inside `ReadLimited`. `ReadLimited` itself should cap growth to `maxBufLen-len(kvb.buf)` before calling slices.Grow.

<h3>5. `Respond` can generate responses that are invalid for the method or status</h3>

`Respond` and `RespondString` always stage Content-Length, write the supplied body, and add Connection: close.

That produces invalid output for:

- Responses to HEAD.
- Informational 1xx responses.
- 204 No Content.
- 304 Not Modified.

Those responses cannot carry a message body; Content-Length also has status-specific restrictions.

The helper should either:

- Enforce the semantics using the request method and status code.
- Be renamed/documented as only suitable for ordinary body-bearing final responses.
- Return an error when given a body for a bodyless response.

</details>

So the tldr is that there are some missed error cases from peers answering unexpectedly again but the happy path should work fine

soypat:
Thank you for the review @MDr164 ! Good catches too. All should be fixed. Will proceed to update cyw43439 and espradio with latest httphi

PR #178, opened 2026-08-10
fix(xnet): make blocking waits deadline-driven instead of iteration-capped

xinix00:
All six StackBlocking wait loops run `for range maxIter` (1000). With a non-sleeping backoff like `BackoffFlagGosched` — the sensible choice on GOMAXPROCS=1 targets, where sleeping starves the NIC poll — those 1000 iterations complete in ~20ms of wall time and silently replace the caller's timeout: a dial with the default 2s timeout fails after ~21ms with a spurious deadline error against any peer slower than that. Observed dialing github.com from a single-core bare-metal (tamago) target through go-net.

The deadline check inside every loop is the real guard, so let it be the only exit and drop the now-unreachable post-loop returns. DoDHCPv4 now checks its deadline unconditionally instead of only on iterations without a state change: with the iteration cap gone, progress-gated checking would let a peer that keeps feeding state transitions hold the loop past any deadline.

The regression test drives a dial through 4000 no-progress wait iterations on simulated time (<5% of the deadline elapsed) and requires establishment once the handshake is finally serviced; on current main it fails at exactly iteration 1000.


soypat:
The reason for maxIter is that it was an easy way to ensure constrained lifetime for the loop. Maybe we can add a condition to the for loop to facilitate the studying of the for loop. i.e: 

```go
ok := true
for ; ok; ok=d.checkDeadline(deadline) {
    // ...
}
return errDeadlineExceeded
```
Let me know what you think! Really exciting to see so many interesting PRs to lneto! 

FYI: currently out of office so I may take some time to get around to the PRs :)

xinix00:
Adopted, thanks. Putting the check in the header also let DoDHCPv4 drop its separate deadline branch, since it now runs every iteration, so the change came out smaller than before.

No rush on the reviews, and enjoy the time off.

x/xnet/stack-blocking.go
```diff
@@ -143,16 +135,16 @@ func (s StackBlocking) DoResolveHardwareAddress6(addr netip.Addr, timeout time.D
...
 		if err == nil {
 			break
-		} else if err = s.checkDeadline(deadline); err != nil {
-			break
 		}
 		s.backoff(backoffs)
 		backoffs++
-		err = errDeadlineExceed // Ensure that if iterations done error is returned.
+	}
+	if err != nil {
```
soypat, line 146: this should probably be err==nil

soypat:
Hey @xinix00 ! I'd really like to get back to your PRs this week before I go on another conference spree next week!

soypat:
Hey! Figured you may not have seen the notifications on your work account so I'm pinging you via this account! @derekdhaas

soypat:
Merged as is with gofix and gofmt applied in a different branch (with your commit) https://github.com/soypat/lneto/pull/198

Thank you for this!

PR #179, opened 2026-08-10
feat(tcp): implement window scaling (RFC 7323)

xinix00:
OptWindowScale is parsed for length validation but never applied, so the effective window is capped by the 16-bit field: throughput tops out at 65535/RTT no matter how large the receive buffer is. A buffer over 64KiB is worse than useless today: the SYN's window wraps in the uint16 conversion, so a 128KiB buffer advertises a near-zero window.

Scaling lives purely at the wire seam in Handler: the ControlBlock keeps real octet counts, conversion happens on frame read (peer windows shifted up, never on SYN segments) and frame write (shifted down, SYN never scaled, saturation instead of wrap). The local shift derives from the receive buffer size in SetBuffers; every active SYN offers the option (a zero shift still lets the peer scale its own window, §2.5) and a SYN-ACK echoes it only when the peer's SYN carried it. The ControlBlock's three 2**16 window caps move to the scaled maximum (65535<<14).

Tests cover on-wire negotiation with asymmetric buffers, a transfer holding >64KiB in flight without a single ACK, and the wire-safety corners including the pre-existing 128KiB SYN wrap. The tcp and xnet fuzzers ran clean for ~21M combined execs. Complements the LossRecovery work (#169): big windows are what make loss recovery worth having.

### DIFF
--- a/tcp/handler.go
+++ b/tcp/handler.go
@@ -29,6 +29,14 @@ type Handler struct {
 	// Read and Write calls belong to the current connection.
 
 	optcodec OptionCodec
+	// Window scaling (RFC 7323 §2). wndShiftLocal is derived from the receive
+	// buffer in [Handler.SetBuffers], wndShiftPeer learned from the peer's offer.
+	// Scaling lives at the wire seam only: the ControlBlock always holds real
+	// octet counts, converted on frame read ([Handler.Recv]) and write
+	// ([Handler.wireWnd]).
+	wndShiftLocal uint8
+	wndShiftPeer  uint8
+	peerOfferedWS bool
 	// reasm tracks out-of-order segments staged in bufRx's free region. Always
 	// enabled once buffers are set (see [Handler.SetBuffers]).
 	reasm reassembly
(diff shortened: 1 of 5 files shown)


soypat:
This is a more far reaching change. Looks overall pretty good but I want to look at this one closely. Next week will check it out closely for sure

MDr164:
Just a heads up, I'm back from traveling now and most of your PRs @xinix00 seem to implement what I already did in my TCP congestion control and loss recovery rework. I initially wanted to prevent others duplicating work I already did but looks like that happened already now. I"ll open a PR from my code as well and then chunk it down again.

soypat:
Will get around to this sometime this week. May need to work it into https://github.com/soypat/lneto/pull/190 as part of tcp.Policy- we'll see. For now thank you for all your contributions @xinix00  :)

tcp/handler.go
soypat, line 471: the changes in wireWnd should probably live within `Handler.recvWindow`. Keep in mind recvWindow is only called in non-SYN situations so you can omit a lot of the logic and checks you've added and simplify code greatly (I suspect).

tcp/handler.go
```diff
@@ -480,6 +498,7 @@ func (h *Handler) Send(b []byte) (int, error) {
 	h.requeueControl = false
 	tfrm.SetSourcePort(h.localPort)
 	tfrm.SetDestinationPort(h.remotePort)
+	segment.WND = h.wireWnd(segment) // wire representation only; scb keeps real octets
```
soypat, line 501: with above change you can do without wireWnd i think

soypat (commented):
On closer inspection I think we can merge this code without needing to include it into `tcp.Policy` interface. This seems like a very constrained improvement. Just need to address some aspects of its current design (opportunity for simplification). @MDr164 would you argue for leaving WS on Policy responsibility?

soypat:
@xinix00 I added your contribution in https://github.com/soypat/lneto/pull/197 ! Thank you a billion for providing this important fix Derek! Closing since this was solved

PR #180, opened 2026-08-10
fix(xnet): allocate ephemeral ports sequentially, not randomly

xinix00:
Every dial draws its ephemeral port at random from the 16384-port dynamic range. Under connection-per-request churn the birthday paradox reuses a recently-released port within a few dozen dials, and the reused 4-tuple lands on whatever state the previous conversation left along the path — the peer's TIME-WAIT socket or a NAT flow-table entry — which silently swallows the new SYN. Observed end to end (HTTP client, keep-alives off, macOS peer): a dead dial after ~286 connections, then ~30s of failing dials — one 2MSL expiry.

Allocate sequentially from a per-stack random offset instead, the way OSes do: a port is only revisited after the full 16384-port cycle, and the random start keeps a rebooted node off the ports its previous life just used.

TestEphemeralPortSequence pins the full-cycle-no-reuse property; TestStackGoTCPDialChurn runs 300 dial/exchange/close rounds over an 8-slot port table as a recycling guard (transient RST refusals are retried — only persistent refusal, a leaked slot, fails). Touches the same test-file tail as #178 — happy to rebase whichever lands second.

### DIFF
--- a/x/xnet/xnet_test.go
+++ b/x/xnet/xnet_test.go
@@ -1196,3 +1196,39 @@ func TestEgressIP_TCPMSSAdvertisesMTU(t *testing.T) {
 		t.Errorf("advertised MSS = %d, want %d (MTU %d - 40)", gotMSS, wantMSS, mtu)
 	}
 }
+
+// TestEphemeralPortSequence checks no ephemeral port is reused before the whole
+// 16384-port dynamic range has cycled, which is what keeps a redial off the
+// teardown state (TIME-WAIT, NAT flow entries) of the conversation before it.
+func TestEphemeralPortSequence(t *testing.T) {
+	s := new(StackAsync)
+	err := s.Reset(StackConfig{
+		Hostname:          "eph",
+		RandSeed:          42,
+		StaticAddress4:    [4]byte{10, 0, 0, 50},
+		MaxActiveTCPPorts: 1,
+		HardwareAddress:   [6]byte{0xbe, 0xef, 0, 0, 0, 50},
+		MTU:               ethernet.MaxMTU,
+		ICMPQueueLimit:    2,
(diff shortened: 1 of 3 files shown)


x/xnet/xnet_test.go
```diff
@@ -1196,3 +1198,211 @@ func TestEgressIP_TCPMSSAdvertisesMTU(t *testing.T) {
...
+			}
+			go func(c net.Conn) {
+				var one [1]byte
+				if _, err := c.Read(one[:]); err == nil {
+					c.Write(one[:])
+				}
+				c.Close()
+			}(c)
+		}
+	}()
```
soypat: Please use the [stackTestScheduler](https://github.com/soypat/lneto/blob/main/internal/ltesto/stackschduler.go) if at all possible. Plenty of examples in xnet tests to see how it is used. Ensure tests do not block/sleep at any point

See https://github.com/soypat/lneto/issues/140

xinix00:
I removed the churn test instead of porting it. It ran three goroutines against the real clock with sleeps and a three-strikes retry tolerance, so it was flaky by construction, which is exactly what #140 is about.

A faithful port is not a straight swap: the test needs a dialer, a server accept loop and a packet pump, and Sched.Goro() panics on the second goroutine ("only one goroutine supported for now"). Happy to port it if Sched grows multi-goroutine support.

What remains is TestEphemeralPortSequence, which pins the property directly with no clock, no sleeping and no goroutines.

soypat:
Check failing lint. likely need to run go fmt or go fix

soypat (approved):
awesome change. Thank you for your contribution!

soypat:
small lint problem, will fix in main

PR #181, opened 2026-08-10
fix(internal): keep the write position when a ring is read empty

xinix00:
`Ring.onReadEnd` and `Ring.ReadDiscard` call `Reset()` when the last readable
byte is consumed, which rewinds `Off` and `End` to 0. That is a valid empty
state, but it also moves the write position — and bytes staged past that
position with `PeekWrite` are addressed relative to it. After the rewind,
`Commit` hands back whatever bytes happen to live at the start of the buffer
instead of the staged ones.

`tcp` stages out-of-order segments exactly this way. From `reassembly.go`:

> No buffer offset is kept: the ring write pointer advances in lockstep with
> rcv.NXT, so the staged bytes are always where seq implies.

Reading is driven by the application, so a receiver that drains its stream
while a gap is open loses the staged tail with no error reported anywhere: the
stream arrives with the correct length and the wrong contents.

### How it showed up

On a Sophgo SG2002 (100Mbit, DWMAC — its receive ring runs dry under load and
drops frames without an error indication, which feeds the receiver a steady
supply of out-of-order segments):

- a 16 MiB HTTP download arrived complete, with a different SHA-256 than the
  server sent;
- TLS over the same path failed with `local error: tls: bad record MAC`;
- the same board's transmit path was bit-perfect across repeated 8 MiB
  transfers, so the corruption is specific to receiving.

With this change the same board completes both, including a 6 MiB HTTPS
transfer that previously died mid-stream.

### The change

Mark the ring empty without rewinding. `End == 0` is what empty means, and the
write position is then `Off` — a state `Ring.Write` already supports explicitly
("Specialised for when user controls Off manually").

### Tests

Both fail before the change:

- `internal.TestRingPeekWriteSurvivesEmptyRead` — stage bytes past the write
  position, drain the ring, fill the gap, commit. On main the read returns
  `"BBBB\x00\x00\x00\x00"` instead of `"BBBBCCCC"`.
- `tcp.TestHandlerStreamIntegrityUnderReorder` — 40 rounds of data segments
  arriving shuffled within blocks of four, with the application draining as it
  goes. On main segment K arrives carrying the contents of an earlier segment.

Full suite green with the change.

### DIFF
--- a/internal/ring.go
+++ b/internal/ring.go
@@ -157,7 +157,7 @@ func (r *Ring) ReadDiscard(n int) error {
 	case n > buffered:
 		return errDiscardExceeds
 	case n == buffered:
-		r.Reset()
+		r.emptied()
 	case n+r.Off > len(r.Buf):
 		r.Off = n - (len(r.Buf) - r.Off)
 	default:
@@ -222,6 +222,18 @@ func (r *Ring) read(b []byte) (n int, err error) {
 func (r *Ring) Reset() {
 	r.Off = 0
 	r.End = 0
+}
+
+// emptied marks the ring empty keeping the write position where it is, unlike
+// [Ring.Reset] which rewinds it to index 0. Bytes staged past that position with
+// [Ring.PeekWrite] are addressed relative to it, so moving it makes a later
+// [Ring.Commit] hand back the wrong bytes.
+func (r *Ring) emptied() {
+	off := r.End
+	if off == len(r.Buf) {
+		off = 0 // Tail exhausted, next write wraps.
+	}
+	r.Off, r.End = off, 0
 }
 
 // Size returns the capacity of the ring buffer.
@@ -306,7 +318,7 @@ func (r *Ring) onReadEnd(totalRead int) {
 	}
 	newOff := r.addOff(r.Off, totalRead)
 	if newOff == r.End {
-		r.Reset()
+		r.emptied()
 	} else if newOff == len(r.Buf) {
 		r.Off = 0 // Optimization case.
 	} else {
--- a/internal/ring_test.go
+++ b/internal/ring_test.go
@@ -661,3 +661,40 @@ func TestRingPeekWriteRejects(t *testing.T) {
 		t.Error("Commit beyond free must error")
 	}
 }
+
+// TestRingPeekWriteSurvivesEmptyRead checks bytes staged with [Ring.PeekWrite]
+// survive the ring being read empty, an event the stager does not control.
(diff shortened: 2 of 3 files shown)


tcp/reorderstream_test.go
```diff
@@ -0,0 +1,142 @@
...
+		return "0"
+	}
+	var d [8]byte
+	i := len(d)
+	for v > 0 {
+		i--
+		d[i] = byte('0' + v%10)
+		v /= 10
+	}
+	return string(d[i:])
```
soypat: strconv.Itoa

internal/ring.go
```diff
@@ -224,6 +224,25 @@ func (r *Ring) Reset() {
...
+// leaving the write position where it is instead of rewinding it to index 0.
+//
+// Both are valid empty states (End==0 is what empty means, and the write
+// position is then Off), but rewinding is only safe when nothing depends on the
+// buffer's physical layout. Bytes staged past the write position with
+// [Ring.PeekWrite] do: they are addressed relative to that position and
+// committed later with [Ring.Commit]. Rewinding would move the position out
+// from under them, so a commit would hand back whatever bytes happen to live at
+// the start of the buffer. TCP stages out-of-order segments this way, which is
+// how a stream can arrive with the correct length and the wrong contents.
```
soypat: No need to restate the entire documentation of how Ring works here. keep comments concise and to the point, no fluff. Careful with claude! Claude has been MUCH more verbose with comments as of late and also much more prone to being incorrect.

internal/ring_test.go
```diff
@@ -661,3 +661,44 @@ func TestRingPeekWriteRejects(t *testing.T) {
 		t.Error("Commit beyond free must error")
 	}
 }
+
+// TestRingPeekWriteSurvivesEmptyRead pins the physical layout that staged bytes
+// depend on: a ring that is read empty must not move its write position, because
+// bytes staged past that position with [Ring.PeekWrite] are addressed relative
+// to it and committed later. Reading is driven by whoever consumes the stream,
+// so the moment the ring goes empty is not under the stager's control.
```
soypat: same for tests. more concise comments

xinix00:
Both done: strconv.Itoa, and the comments are trimmed in the code and in the tests. Fair point on the `emptied` doc, it restated Ring at length; it is four lines now and says only how it differs from Reset.

soypat (approved):

soypat:
Looks great. If you can please fix the CI/lint error. You can run `go fix ./...` in lneto repo and push changes.

soypat:
same as other PR, will fix lint error in main

PR #182, opened 2026-08-10
fix(tcp): retransmit unacknowledged data after a local close

xinix00:
`ControlBlock.Send` refused any outgoing segment carrying data in FIN-WAIT-1,
citing RFC 9293's "no further SENDs from the user will be accepted by the TCP
implementation". That rule bounds what the application may queue —
`Handler.Write` already enforces it — and not the retransmission of data the
connection has already accepted from it.

The consequence is that write-then-close, which is what nearly every server does
with a response, cannot recover from losing its last data segment. The FIN
occupies a sequence number above that data, so the peer cannot cross the gap to
process the close: it waits for bytes that are never resent while the sender
waits for an ACK that cannot arrive. Neither side times out at the TCP layer.

Two details suggest this was unintended rather than deliberate:

- the restriction lived in `ControlBlock.Send` and in `PendingSegment` with
  different state sets — `Send` rejected data in FIN-WAIT-1 and FIN-WAIT-2 while
  letting CLOSING and LAST-ACK through, `PendingSegment` refused all four;
- the rule it cites is already enforced where it belongs, in `Handler.Write`.

FIN-WAIT-2 keeps the restriction and gains the reasoning: it is reached by our
FIN being acknowledged, which acknowledges everything below it, so no
unacknowledged data can remain there.

`PendingSegment` needs the same distinction, since it decides whether to offer
send-buffer data at all. It gets an unexported `State` predicate rather than a
new exported one; both halves are required — with only the `Send` validation
relaxed the test below still fails, since no data is offered in the first place.

### Tests

- `TestHandlerRetransmitsAfterRTO` covers the `Handler`/`LossRecovery` seam that
  the RTO unit tests do not reach: they exercise the state machine in isolation,
  where it behaves correctly. Passes before and after; it is the guard for the
  seam itself.
- `TestHandlerRetransmitsAfterCloseWithUnackedData` is red before the change
  ("no retransmission in FIN-WAIT-1 — unacknowledged data is stranded by the
  close") and green after.

Both use an injected clock, so they are deterministic and take no wall time.
Full suite green.

### DIFF
--- a/tcp/control.go
+++ b/tcp/control.go
@@ -279,10 +279,9 @@ func (tcb *ControlBlock) PendingSegment(payloadLen int) (_ Segment, ok bool) {
 		// Optimist Strategy: retransmit oldest data once.
 		return Segment{SEQ: tcb.snd.UNA, DATALEN: Size(payloadLen), ACK: tcb.rcv.NXT, WND: tcb.rcv.WND, Flags: FlagACK}, true
 	}
-	established := tcb._state == StateEstablished
-	canSendData := established || tcb._state == StateCloseWait
+	canSendData := tcb._state.txQueuedDataOpen()
 	if !canSendData {
-		payloadLen = 0 // Can't send data if not established or close-wait.
+		payloadLen = 0 // No send-buffer data may go out in this state.
 	}
 	if pending == 0 && payloadLen == 0 {
 		return Segment{}, false // No pending segment.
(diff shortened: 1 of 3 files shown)


tcp/definitions.go
```diff
@@ -329,6 +329,16 @@ func (s State) TxDataOpen() bool {
 	return s == StateEstablished || s == StateCloseWait
 }
 
+// TxQueuedDataOpen returns true if the state may still put already-queued
+// send-buffer data on the wire. It differs from [State.TxDataOpen], which answers
+// whether the application may queue more: after a local close the FIN occupies a
+// sequence number above data the peer has not acknowledged, so that data must
+// still be transmitted (or retransmitted) or neither side can make progress —
+// the peer cannot process the FIN across a gap. RFC 9293 §3.10.8.
```
soypat: The wording is confusing. Is 87% of the documentation of this function how TxDataOpen (a separate func) works?

Refer to my comment on LLM comments in the previous PR.

tcp/control.go
```diff
@@ -522,8 +521,16 @@ func (tcb *ControlBlock) validateOutgoingSegment(seg Segment) (err error) {
...
-		err = errConnectionClosing // Case 1: No further SENDs from the user will be accepted by the TCP implementation.
+	case seg.DATALEN > 0 && tcb._state == StateFinWait2:
+		// FIN-WAIT-2 is reached by our FIN being acknowledged, which acknowledges
+		// everything below it: no unacknowledged data can remain, so data here is
+		// a caller error. FIN-WAIT-1 is different — the FIN sits above data the
+		// peer may not have received yet, and that data must still go out or the
+		// peer can never cross the gap to process the close (RFC 9293 §3.10.8).
+		// The rule that "no further SENDs from the user will be accepted" bounds
+		// what the application may queue, which [Handler.Write] enforces; it does
+		// not bound retransmission of data already accepted from it.
```
soypat: Refer to AI comment comment

soypat (commented):
Refer to https://github.com/soypat/lneto/pull/181#discussion_r3754480546

xinix00:
Rewritten. The doc now describes the predicate itself instead of TxDataOpen, and the method is unexported: the send path needs the distinction internally, but the fix does not need a new public method on State. The FIN-WAIT-2 comment keeps only why FIN-WAIT-1 is excluded, with the section reference.

soypat (approved):
looks excellent. Thank you for your contribution!

PR #183, opened 2026-08-10
fix(xnet): install a retransmission timer on TCP connections

xinix00:
No connection created through `x/xnet` had one. Neither `NewTCPPool` nor the dial
path in `StackGo` set `ConnConfig.LossRecovery` and `ConnConfig.Nanotime`, so
`tcp.Handler` ran with loss recovery disabled: nothing noticed a lost segment,
and the connection simply stopped — the sender waiting for an ACK that cannot
arrive, the receiver for data nobody will resend.

`TCPPoolConfig.NanoTime` already documents itself as:

> NanoTime returns the current monotonic time in nanoseconds. Used for pool
> timeout tracking **and passed to each [tcp.Conn] for retransmission timing
> (RFC 6298)**. If nil, defaults to time.Now().UnixNano().

This makes that true. Each connection gets its own `tcp.RTO`, which shadows that
connection's send sequence space and so cannot be shared: 80 bytes per
connection, allocated once at pool construction, against per-connection buffers
measured in kilobytes. The clock comes from `TCPPool.now`, keeping the existing
`NanoTime` override and its `time.Now` fallback.

### Measurement

A 512 KiB stream through an in-process pump that drops every 23rd data frame:
**2 of 10 runs completed before, 8 of 10 after** (the residual failure is a
separate and rarer stall, still under investigation — not addressed here).

### Test

`TestTCPRetransmitsLostSegment` drops exactly one data segment and requires the
byte to arrive anyway. It depends on #182, since the server closes after writing
and the retransmission then has to leave FIN-WAIT-1.

### If you would rather keep this opt-in

A connection without a retransmission timer cannot recover from any loss, which
is why this wires it unconditionally rather than adding a config knob. But the
call is yours: the alternative is to correct `NanoTime`'s documentation and
expose a hook for callers to supply their own `LossRecovery`. Happy to do it that
way instead — say which you prefer.

### DIFF
--- a/tcp/control.go
+++ b/tcp/control.go
@@ -279,10 +279,9 @@ func (tcb *ControlBlock) PendingSegment(payloadLen int) (_ Segment, ok bool) {
 		// Optimist Strategy: retransmit oldest data once.
 		return Segment{SEQ: tcb.snd.UNA, DATALEN: Size(payloadLen), ACK: tcb.rcv.NXT, WND: tcb.rcv.WND, Flags: FlagACK}, true
 	}
-	established := tcb._state == StateEstablished
-	canSendData := established || tcb._state == StateCloseWait
+	canSendData := tcb._state.txQueuedDataOpen()
 	if !canSendData {
-		payloadLen = 0 // Can't send data if not established or close-wait.
+		payloadLen = 0 // No send-buffer data may go out in this state.
 	}
 	if pending == 0 && payloadLen == 0 {
 		return Segment{}, false // No pending segment.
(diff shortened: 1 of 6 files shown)


soypat:
I'll review this after #182

xinix00:
Rebased on the updated #182. Worth flagging before you get to it: the test here is wall-clock dependent too, since it waits out a real RTO, and it hits the same Sched limitation as #180 (the pump and the server each need a goroutine). Say the word if you want it reshaped first.

soypat:
So the new breaking `Policy` (#190 ) change may need some rethinking of how we include retransmission timers and may require some rethinking of this and #182. We'll get there :)

soypat:
Closing in favor of https://github.com/soypat/lneto/pull/199/changes which rebases these changes and fixes race conditions. Will keep #199  one open for a bit for review.

PR #186, opened 2026-08-15
fix(dns): decode up to four answer records

hnw:
## Summary

Allow `dns.Client` to retain up to four answer records from a response instead of using the number of questions as the answer decode limit.

Previously, `Client.StartResolve` used the question count as both the question and answer decode limits. Lookups initiated through `StackRetrying.DoLookupIP` send one DNS question, so a response containing multiple A records exposed only the first address.

## Motivation

I noticed this while resolving `pool.ntp.org` through `StackRetrying.DoLookupIP` in an application using lneto.

The DNS response contained multiple A records, but `DoLookupIP` returned only one address. If that address did not respond, there were no alternative addresses available for fallback.

After increasing the answer decode limit to four, the same lookup returned four addresses.

## Rationale

`StackAsync.StartLookupIPType`, the only production caller of `Client.StartResolve` in this repository, always sends exactly one question.

Direct callers can currently provide more than one question, but `Client.StartResolve` generates standard DNS QUERY messages (`OPCODE=0`). [RFC 9619](https://www.rfc-editor.org/rfc/rfc9619#section-4) prohibits those messages from containing more than one question. Therefore, replacing the question-derived answer limit with four does not reduce RFC-compliant behavior.

The limit of four matches the `[4]netip.Addr` lookup-result buffer used by `x/xnet.StackAsync`. Records beyond this limit are still skipped while parsing the response.

## Testing

`TestClient_ReceivesDNSResponse` now covers three response shapes:

- One answer, preserving the existing client behavior.
- Four answers, covering the reported regression.
- Five answers, verifying that the first four are retained and the
  configured limit is enforced.

Each case verifies both address lookup and response copying.

### DIFF
--- a/dns/client.go
+++ b/dns/client.go
@@ -38,7 +38,7 @@ func (c *Client) StartResolve(localPort, txid uint16, cfg ResolveConfig) error {
 		return lneto.ErrInvalidConfig
 	}
 	c.reset(localPort, txid, CQueryPending, cfg.EnableRecursion)
-	c.msg.LimitResourceDecoding(uint16(nd), uint16(nd), 0, 0)
+	c.msg.LimitResourceDecoding(uint16(nd), 4, 0, 0)
 	c.msg.AddQuestions(cfg.Questions)
 	c.msg.AddAdditionals(cfg.Additional)
 	return nil
--- a/dns/client.go
+++ b/dns/client.go
@@ -24,6 +24,9 @@ type ResolveConfig struct {
 	Questions       []Question
 	Additional      []Resource
 	EnableRecursion bool
+	// MaxResponseAnswers limits how many answer records are decoded from the
+	// DNS response. If zero it defaults to the number of Questions.
+	MaxResponseAnswers uint16
 }
 
 func (sudp *Client) Protocol() uint64 { return uint64(lneto.IPProtoUDP) }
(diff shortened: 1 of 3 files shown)


dns/client.go
```diff
@@ -38,7 +38,7 @@ func (c *Client) StartResolve(localPort, txid uint16, cfg ResolveConfig) error {
 		return lneto.ErrInvalidConfig
 	}
 	c.reset(localPort, txid, CQueryPending, cfg.EnableRecursion)
-	c.msg.LimitResourceDecoding(uint16(nd), uint16(nd), 0, 0)
```
soypat, line 41: Maybe better to then use `max(uint16(nd), 4)` which would also not limit number of answers if there are many questions. Alternatively (and probably even better) have ResolveConfig declare the number of answers. I think this'd be the best way forward. We then have to update all call sites to use new field.

soypat (changes requested):

dns/client.go
hnw: Thanks for the review and the excellent suggestion!
I completely agree that having `ResolveConfig` declare the number of answers is the best approach. It keeps the `dns` package flexible without hardcoding any limits.
I've just pushed an update that:
1. Adds `MaxResponseAnswers` to `dns.ResolveConfig` (defaulting to the number of questions if left as zero to preserve backward compatibility).
2. Updates `x/xnet.StackAsync` to set this limit dynamically based on its buffer size (`uint16(len(s.addrbufnip))`).
3. Updates the relevant tests.


Let me know if this looks good to you!

soypat:
@hnw Thank you for your contribution! Good stuff :)

PR #187, opened 2026-08-17
TCP refactor [DO-NOT-MERGE]

MDr164:
To increase visibility I"m moving this branch/PR over here. So far I have not run into any larger issues. I"ll gradually split it up into standalone fixes, low stake additions and also the more 'controversial' interface changes. The contents of this branch have largely been discussed in other issues, PRs, threads etc but as the info has been scattered other people started working on the very same items already done in this branch e.g. at #179


soypat:
If you have tinygo installed I'd be curious about running the README's table generation for evaluating binary size increase in this PR alone

MDr164:
I rebuilt the example from the README with the current reference being:
amd64 1.6 MB, WASM 1.2 MB, Pico 192 kB

And freshly build with TinyGo 0.41.1:
amd64 1.6 MB, WASM 1.2 MB, Pico 187 kB

So DCE is working fine and we're not pulling anything in that we really don't need :)
I got the same result building this branch and main as all features here are truly optional and not bloating a default build.

MDr164:
I think that it's a good idea to update the benchmark bot comment to show a diff between what's currently in main vs the PR branch and add size to it while we're at it. That's a lot more valuable than just looking at the branch results alone imo.

soypat:
We can have the benchmark conversation in an issue or the other PR!

soypat:
DCE was working too well in minimal-working-example. The binary increase is not accurate!

https://github.com/soypat/lneto/pull/196

MDr164:
I'll rebase this here against all the changes you already merged to main, that should then also kick off the memci task in here :+1:

MDr164:
Now with memci actually showing us the diff it is fairly small with a 3% increase for the pico binary. It is however noticeable though. The regular Go build aka the one I usually care about is increased by less than 1%

MDr164:
So the biggest change is the deadline handling. If you'd like I can already slim this PR down and try out GitHubs new stacked PRs feature which should be available now.

MDr164:
Because things around the deadline/time work got a bit scattered I asked my clanker for an overview of where we're at right now:

<details>
<summary>Overview</summary>

### What the recent PRs did
- **#198** (from #178): blocking waits are now bounded by a *deadline* via an injectable `StackBlocking._nanotime` (was `for range maxIter(1000)`, which silently replaced the caller's timeout on non-sleeping backoffs).
- **#199** (from #183): pool/dial connections now install an `rto.Timer` per connection. That PR also flagged a *"separate and rarer stall, still under investigation."*
- **#200**: `TCPListener` closed-bool race; not timing-related.

### The actual problem: three clocks, not unified
| Concern | Clock used today | In `NextDeadline()`? |
|---|---|---|
| RTO / retransmission (`Policy`) | injected `ConnConfig.Nanotime` (`func() int64`) | yes |
| Blocking-wait timeouts (#198) | injected `StackBlocking._nanotime` | no (spins to deadline) |
| **I/O deadlines** (`Conn.SetReadDeadline`/`SetWriteDeadline`) | **hardcoded wall clock** (`rdead`/`wdead` via `time.Since`) | **no** |
| 2MSL TIME-WAIT | **no timer** — `StateTimeWait` is entered but never left by time | no |
| Keepalive | primitives exist (`MakeKeepalive`) but **no timer drives them** | no |

The aggregation plumbing is already there: `StackNode.NextDeadline()` is interface-level (with `lneto.NoDeadline` for nodes without timing) and bubbles up `Handler → ip → StackAsync.NextDeadline()`. But `Handler.NextDeadline()` only returns the Policy's RTO deadline today. Its own TODO says it best:

> only the installed policy feeds this deadline. The 2MSL TIME-WAIT timer and the keepalive interval are also time-driven … They should be folded in here as the earliest of all pending deadlines **once the `Conn.SetDeadline` family is reconciled with `ConnConfig.Nanotime`**.

So the crux: **the `net.Conn` I/O-deadline API runs off `time.Time`/`time.Since` (wall clock), while everything we actually schedule against runs off an injected monotonic `func() int64`.** On a simulated-time or bare-metal custom-clock target — the whole reason `Nanotime` exists — I/O deadlines diverge from RTO/TIME-WAIT/keepalive and can't be driven or tested off the same clock.

### The decision
Adopt one monotonic source (`ConnConfig.Nanotime`) for **all** time-driven TCP behavior, then:
1. **Reconcile `Conn.SetDeadline`** to compare `rdead`/`wdead` against `Nanotime()` instead of `time.Since` (this is the `net.Conn` seam, i.e. the "controversial interface change" flagged in #187).
2. **Fold all deadlines into `Handler.NextDeadline()`** (earliest of {RTO, 2MSL, keepalive, pending read/write deadline}) so an event loop wakes on `StackAsync.NextDeadline()` rather than spinning.
3. **Add the missing timers** (2MSL TIME-WAIT exit, keepalive) as core `ControlBlock` deadlines rather than another `Policy`, since they must work with no policy installed.
4. Optionally converge `StackBlocking._nanotime` (and `DoPing`, which still uses `time.Now`) onto the same source.

</details>

soypat:
I'm gonna try to sit down today see what I make of these. Thank you for the heads up! I was also wondering where we were standing after all those merges. After today I will have to sit down to work on my talk for Fyne Conf on friday though. Wishing you the best with your talk preparations Marvin!

MDr164:
I'll try githubs stacked PRs feature with this commit collection here, so prepare for some noise. The next PRs are essentially this one here but reviewable.

tcp/policy.go
```diff
@@ -1,33 +1,313 @@
...
+	// this same directive, pending control segments and ACKs still proceed; only
+	// fresh data from the send buffer is withheld until a later PreTx clears it.
+	HoldNew bool
+}
+
+// RxEvent reports what a connection did with a received segment. It is passed by
+// value to [Policy.PostRx] and must not be retained.
+//
+// It exists so that a policy does not have to reconstruct the state machine's
+// decisions from the raw segment. Reconstructing them means duplicating the
```
soypat, line 268: I'm confused as to what this means "duplicating acceptance rules". It seems it is justifying divergence from current design but the justification is not clear to me

MDr164: By "acceptance rules" I mean TCP's own checks for whether a received segment counts, iirc the RFC 9293 §3.4 sequence/window calls it acceptability tests, whether the ACK is in range, whether it's a duplicate or out-of-order, etc. The connection already runs these to decide what it does with the segment. Should I change the wording of this?

soypat: OK, I follow- but I believe if the worry is that the logic is duplicated then we can extract the check to it's own function so it may be used wherever by whomever, no duping. I will continue looking at the rest of the PR for now, but I'm pretty sure we can omit a lot of the type definitions with the current design by being a little creative.

MDr164: Feel free to look at the smaller PR instead of this one as it's vastly similar (I actually reorganized some commits there to make it work in a smaller condensed PR). I'll likely not touch this branch here soon as it has gotten too large.

PR #188, opened 2026-08-18
CI benchmark and size comparison against main

MDr164:
This change allows the github action bot comment to present more meaningful data instead of just capturing the current branch. I can drop the precommit hook commit from the commits in this PR, just had it around and I recall you wanted to take a look at that as well @soypat


.github/workflows/ci.yaml
```diff
@@ -133,5 +161,40 @@ jobs:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
+          path: main
+
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version: "1.26"
+
+      - uses: acifani/setup-tinygo@dd8a7075d951a7595b2ef2123ed0ab1af0c13e56 # v3.0.0
```
MDr164, line 164: Just letting you know that I used this one again here for the tinygo stuff, we haven't decided yet if/how we want to introduce tinygo to CI so I just left it in here for reference, I can drop it easily as it's in a separate commit.

soypat:
So this seems like an god awful lot of code for just adding a new table with a few rows to the existing github comment. Not worth merging as-is imho. I have nothing against including acifani's CI step. Checked users of it and recognized a bunch of respectable peeps in there.

MDr164:
Reduced it to just the extra comparison against main and kept the other two commits locally so I can open separate PRs for them

soypat (commented):
I'm wondering if it makes sense to use the github artifact upload/download actions to store benchmark results or if the run both on every PR makes more sense. 
pros for artifacts is that it may simplify the CI.yaml.
Downsides is that github worker availability quality may vary over time, which it already seems to vary for the workflow, and quite a bit more than what I'd hope. To have a benchmark show a diff of 5% makes me question the value of this PR. Furthermore I have yet to see a performance PR that would put this to good use. This is not a permanent "No" though. github action runners may be shit for benchmarks but there are other ways... I'm thinking about TinyGo HCI...

MDr164:
> Github worker availability quality may vary over time

That and artifacts can also only be held for around 90 days max unless they are stored in releases or packages, so by using the artifacts for more longterm comparison then we would need to rerun it every now and then either way

> Which already seems to vary for the workflow, and quite a bit more than what I'd hope.

True, the compute has quite a lot of variance in it but the same is true for the current state, I now made it visible. It's probably more worthwhile for allocations and memory only as that's more predictable

> To have a benchmark show a diff of 5% makes me question the value of this PR

I wouldn't question the value of this PR, I would more so question the value of displaying the ns/op metric at all. E.g. drop it from the table and just look at B/op and allocs/op only. Or even better just have the bot throw a warning text once something that shouldn't allocate starts to do it (e.g. check for non-zero)

> Furthermore I have yet to see a performance PR that would put this to good use.

Fair, it's less about the actual benchmark but more so about putting the values the github bot posts into perspective as to me they lacked context so running them against main did just that.

> I'm thinking about TinyGo HCI...

That would be more valuable for performance counters like ns/op for sure but increases the maintenance burden by a bit having to put a small test rig into the network to test on an actual microcontroller.

soypat:
Sorry for delay once more, I've arrived back home yesterday and still getting settled. yeah I'm iffy on ns/op unless we can lower the jitter. A bot that only comments when bytes/op is non-zero sounds really cool... now i think of it though this could be so much simpler: just capture the benchmark output in CI and parse it and return non-zero output if anything allocates ?

Yeah the HCI is far in the future. just daydreaming

MDr164:
> [...] Unless we can lower the jitter

We're running on colocated VMs in the CI, I highly doubt that there is a decent way of achieving that with GitHubs runners tbh

> [...] Just capture the benchmark output in CI and parse it and return non-zero output if anything allocates?

That is roughly what I had in mind as well, yes. Want me to rework it like that?

> Yeah the HCI is far in the future

Well we do run hardware at work connected to GitHub CI so I kinda know how it's done but someone will need to maintain it after the initial setup which is the larger item.

soypat:
> That is roughly what I had in mind as well, yes. Want me to rework it like that?

Yeah! Let's give it a try. Should result in deleting code/config which is something I love.

soypat:
@MDr164 I just realized! You know what else is jitter free? *Binary size*. That'd be worth keeping the comment for.

MDr164:
Pushed another commit (now deleting more than adding in this PR) where most things now happen in yaml as it could make it simpler. Not 100% happy but also having a full Go helper tool is also not warranted anymore with those changes. What do you think?

.github/workflows/ci.yaml
soypat: We should do benchmarks+binary size with this image so we can track heap allocs and binary size with both tinygo builds and normal go builds. Some patterns that do not allocate in normal Go do allocate in tinygo and vice versa. This would be extremely valuable to have

soypat:
I've left a comment in one of your review line-comments: basically I think we could do away with ns/op and do binary-size and heap allocs measured with both tinygo and normal Go since their heap allocation escapes differ. I forget if we currently use a Go program to parse the benchmarks, if we do we can add the logic there and separate between Go/tinygo to generate the comment contents. I'm particularily fond of the pattern we use in https://github.com/tinygo-org/tinybench for doing this. Very data driven design

soypat:
I went on a sidequest to add DWARF to my tinyboot repo and create a binary-size measureing tool that assigns size increases on a per package basis. Below is an older version of main branch from a few weeks ago compared to after the latest DNS merge.

```
go install github.com/soypat/tinyboot/cmd/bindiff@latest
bindiff -mem -kind package diff min-working-example-old min-working-example
KIND     NAME                                    OLD        NEW      DELTA         
package  [unattributed]                      1373085    1373524       +439    +0.0%
package  github.com/soypat/lneto/dns            8611       8798       +187    +2.2%
package  go:func                              201832     201880        +48    +0.0%
package  github.com/soypat/lneto/internal      10783      10820        +37    +0.3%
package  github.com/soypat/lneto/x/xnet        36209      36238        +29    +0.1%
package  github.com/soypat/lneto/tcp           61461      61473        +12    +0.0%
package  runtime                              694425     694426         +1    +0.0%
TOTAL                                        3019048    3019801       +753    +0.0%
```
I think this would be a excellent tool to `go install` and use in CI to print out differences with tinygo builds and normal Go builds. What do you think?


FYI unattributed breaks down into several parts
```
bindiff -mem -kind symbol profile min-working-example | grep unattr   
symbol   [unattributed] .gopclntab                                        909840
symbol   [unattributed] .rodata                                           369473
symbol   [unattributed] .text                                              49815
symbol   [unattributed] .noptrdata                                         25336
symbol   [unattributed] .data                                              13952
symbol   [unattributed] .dynsym                                              936
symbol   [unattributed] .rela.plt                                            816
symbol   [unattributed] .bss                                                 616
symbol   [unattributed] .dynstr                                              601
symbol   [unattributed] .plt                                                 560
symbol   [unattributed] .noptrbss                                            409
symbol   [unattributed] .dynamic                                             304
symbol   [unattributed] .got.plt                                             296
symbol   [unattributed] .hash                                                192
symbol   [unattributed] .note.go.buildid                                     100
symbol   [unattributed] .gnu.version_r                                        96
symbol   [unattributed] .gnu.version                                          78
symbol   [unattributed] .note.gnu.build-id                                    36
symbol   [unattributed] .interp                                               28
symbol   [unattributed] .rela                                                 24
symbol   [unattributed] .got                                                   8
symbol   [unattributed] .tbss                                                  8
```

MDr164:
Looks like a nice output to have in CI for sure. I'll try to revisit all the PRs and changes I have on this repo this week. Doing a bit of conference prep again which is eating more time than anticipated.

MDr164:
I see you factored it out to https://github.com/soypat/memci so I'll do my best today to work through my open items here

soypat:
Yes! I think it looks good by me. Let me know if you think it is missing something
https://github.com/soypat/tinyboot/pull/3#issuecomment-5509307316

soypat:
FYI, https://github.com/soypat/lneto/pull/196 has memci built in as of now, you can checek the output here for that PR: https://github.com/soypat/lneto/actions/runs/33763285799/attempts/1#summary-100675060664

soypat:
memci seems to have superceded this. Thanks for the inspiration @MDr164 :)

PR #189, opened 2026-08-22
fix(dns): resolve CNAME chains, eagerly decode CNAME target into r.data

hnw:
## Summary

Fixes an issue where `DoLookupIP` returned corrupted IPv6 addresses or failed with `ErrInvalidAddr` when querying hostnames that return CNAME records (e.g., `www.yahoo.co.jp`, `www.bing.com`).

## Root Cause

`Message.WriteAnswers` previously attempted to decode the raw RDATA of all matching answer records into IP addresses (`netip.AddrFromSlice`) without validating the record type. As a result, CNAME target domain strings were incorrectly interpreted as IPv6 addresses or failed length checks (`ErrInvalidAddr`).

## Changes

- **CNAME Chain Resolution**:
  - `Message.WriteAnswers` follows in-band CNAME alias chains and resolves the canonical name's A/AAAA IP addresses.
  - Bounded CNAME chain resolution to safely handle cyclic or malformed alias chains without infinite loops.
- **Resource decoding** (per review discussion):
  - `Resource.Decode` expands CNAME RDATA in place into `r.data` (uncompressed wire format) while the full message is still available, and updates `header.Length` accordingly — the name bytes are stored exactly once.
  - Added `Resource.CNAMEView()` returning a length-bounded view of the expanded target name; `Message.WriteAnswers` uses it internally.
- **Configuration Limits**:
  - Replaced `MaxResponseAnswers` with `MaxIPs` and `MaxCNAMEs` in `ResolveConfig`. `MaxCNAMEs` is strictly explicit — no hidden default inside `Client.StartResolve`; callers opt in (`x/xnet` sets 8).
- **Tests**:
  - Regression tests only: the www.yahoo.co.jp case (compressed CNAME RDATA), reordered answers, and cyclic CNAME chains.

## Verification

- Ran unit tests via `go test ./dns/...` and `go test ./x/xnet/...` to confirm all new and existing tests pass.

### DIFF
--- a/dns/client.go
+++ b/dns/client.go
@@ -43,11 +43,7 @@ func (c *Client) StartResolve(localPort, txid uint16, cfg ResolveConfig) error {
 	if maxIPs == 0 {
 		maxIPs = uint16(nd)
 	}
-	maxCNAMEs := cfg.MaxCNAMEs
-	if maxCNAMEs == 0 {
-		maxCNAMEs = 16
-	}
-	maxAns := maxIPs + maxCNAMEs
+	maxAns := maxIPs + cfg.MaxCNAMEs
 	c.reset(localPort, txid, CQueryPending, cfg.EnableRecursion)
 	c.msg.LimitResourceDecoding(uint16(nd), maxAns, 0, 0)
 	c.msg.AddQuestions(cfg.Questions)
--- a/dns/client.go
+++ b/dns/client.go
@@ -24,9 +24,8 @@ type ResolveConfig struct {
 	Questions       []Question
 	Additional      []Resource
 	EnableRecursion bool
-	// MaxResponseAnswers limits how many answer records are decoded from the
-	// DNS response. If zero it defaults to the number of Questions.
-	MaxResponseAnswers uint16
+	MaxIPs          uint16
+	MaxCNAMEs       uint16
 }
 
 func (sudp *Client) Protocol() uint64 { return uint64(lneto.IPProtoUDP) }
@@ -37,13 +36,18 @@ func (sudp *Client) ConnectionID() *uint64 { return &sudp.connID }
 
 func (c *Client) StartResolve(localPort, txid uint16, cfg ResolveConfig) error {
 	nd := len(cfg.Questions)
-	if nd > math.MaxUint16 {
+	if nd > math.MaxUint16 || nd == 0 {
 		return lneto.ErrInvalidConfig
 	}
-	maxAns := cfg.MaxResponseAnswers
-	if maxAns == 0 {
-		maxAns = uint16(nd)
+	maxIPs := cfg.MaxIPs
+	if maxIPs == 0 {
+		maxIPs = uint16(nd)
 	}
+	maxCNAMEs := cfg.MaxCNAMEs
+	if maxCNAMEs == 0 {
+		maxCNAMEs = 16
+	}
+	maxAns := maxIPs + maxCNAMEs
 	c.reset(localPort, txid, CQueryPending, cfg.EnableRecursion)
 	c.msg.LimitResourceDecoding(uint16(nd), maxAns, 0, 0)
 	c.msg.AddQuestions(cfg.Questions)
--- a/dns/dns.go
+++ b/dns/dns.go
@@ -48,6 +48,7 @@ type Question struct {
 type Resource struct {
 	header ResourceHeader
 	data   []byte
+	target Name
 }
 
 // A ResourceHeader is the header of a DNS resource record. There are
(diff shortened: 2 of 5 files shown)


dns/dns.go
```diff
@@ -460,6 +485,14 @@ func (r *Resource) Decode(b []byte, off uint16) (uint16, error) {
...
 		return off, errResourceLen
 	}
+	if r.header.Type == TypeCNAME {
+		_, err = r.target.Decode(b, off)
+		if err != nil {
+			r.target.Reset() // Tolerate undecodable target; CNAME chain lookup skips it.
+		}
+	} else {
+		r.target.Reset()
+	}
```
soypat, line 520: IT seems this duplicates memory. We already have the name bytes in the data buffer, We can have a method that is `CopyName(*Name) error` on Resource for CNAMEs that decodes the name data so we don't need to eagerly do it. This would allow us to drop the target field which is redundant with data

dns/dns.go
```diff
@@ -300,22 +301,45 @@ func (m *Message) AppendTo(buf []byte, txid uint16, flags HeaderFlags) (_ []byte
...
-			continue
+	var alias Name
+	// Each pass follows at most one alias hop
+	// The pass limit also bounds CNAME cycles in malformed responses.
+	for range m.Answers {
+		var next Name
+		n = 0
+		for i := range m.Answers {
+			if int(n) >= len(dst) {
+				break
```
soypat: This dropped the Exhausted error which is useful signalling.

dns/dns.go
```diff
@@ -300,22 +301,45 @@ func (m *Message) AppendTo(buf []byte, txid uint16, flags HeaderFlags) (_ []byte
...
-		hdr := ans.Header()
-		if !hdr.Name.EqualString(host) {
-			continue
+	var alias Name
+	// Each pass follows at most one alias hop
+	// The pass limit also bounds CNAME cycles in malformed responses.
+	for range m.Answers {
+		var next Name
+		n = 0
+		for i := range m.Answers {
```
soypat, line 311: why are you n^2 looping? I think this is most likely a mistake on your part?

dns/client.go
```diff
@@ -37,13 +36,18 @@ func (sudp *Client) ConnectionID() *uint64 { return &sudp.connID }
...
-	if maxAns == 0 {
-		maxAns = uint16(nd)
+	maxIPs := cfg.MaxIPs
+	if maxIPs == 0 {
+		maxIPs = uint16(nd)
 	}
+	maxCNAMEs := cfg.MaxCNAMEs
+	if maxCNAMEs == 0 {
+		maxCNAMEs = 16
+	}
```
soypat: Same logic as last PR, why did you back off from using ResolveConfig????

dns/dns_test.go
```diff
@@ -239,26 +239,178 @@ func TestDecodeMessage(t *testing.T) {
...
+	}
+}
+
 func TestClient_ReceivesDNSResponse(t *testing.T) {
 	const hostname = "example.com"
 	const txid = uint16(12345)
 	const clientPort = uint16(54321)
-	const maxAnswers = 4
-	allIPs := [5][4]byte{
+	const maxIPs = 4
```
soypat, line 375: I have not tested yet, but ensure that your tests are regression tests only. I will ask you top revert any non-regression test or superfluous tests. I'm not a fan of LLMs adding a slew of tests every PR for small changes

soypat (changes requested):

dns/dns.go
hnw: That's a great point, thanks!

I was originally trying to keep raw wire data (including compression pointers) inside `r.data`. Thanks to your feedback, I realized we can simply write the decoded target name directly into `r.data` instead.

I'll go ahead and remove `r.target` and update `Resource.Decode` to write the decoded name into `r.data` (for CNAMEs). This completely eliminates the memory duplication while keeping the `Resource` struct minimal.

hnw: Ah, I misread your suggestion — you meant lazy decoding via `Resource.CopyName()`, but I went ahead with eager decoding into `r.data`. Sorry about that.

One thing about `CopyName`: resolving compression pointers requires the message in its original contiguous form, since pointers reference absolute offsets into it. `Decode` copies each record's RDATA into its own `Resource` and drops the reference to the full buffer on return, so there's nothing left to resolve pointers against by the time `CopyName` could run.

So the two realistic directions are:

- Eager: decode in `Resource.Decode` while the message is still available. No duplication, no extra field; cost is that CNAME `r.data` holds the uncompressed name rather than the wire bytes.
- Lazy: keep `r.data` strictly raw and retain a slice of the original msg somewhere (on `Message` or each `Resource`) so `CopyName(*Name)` can resolve pointers on demand.

I leaned toward the former given the lifetime issue, but happy to go the other way if you'd prefer keeping `r.data` raw. What do you think?

soypat: Hmm yeah, eager sounds better. Sorry about mixing things up. We certainly don't want to keep references to the message. So this path would mean that if the resource is a CNAME the following could be a valid approach:

```go
func (rsc *Resource) CNAMEView() (Name, bool) {
   if rsc.header.Type == TypeCNAME {
       return Name{data: rsc.data}
   }
   return Name{}, false 
}
```
Is this aligned with your interpretation of eager?

hnw: Yes, exactly aligned. Implemented in 8c49548:

- `Resource.Decode` expands CNAME RDATA in place into `r.data` (uncompressed wire format) while the full message is still available, and adjusts `header.Length` to match.
- The accessor landed as `CNAMEView()`, built from `RawData()` so the view can never expose more bytes than `header.Length` claims.

`Message.WriteAnswers` now uses it internally instead of doing its own aliasing.

dns/dns.go
hnw: Restored in the previous revision — `ErrExhausted` is returned when `dst` fills up after copying the available addresses.

dns/dns.go
hnw: Not a mistake — the double scan is what resolving in-band CNAME chains requires when answer order isn't guaranteed: each hop can sit anywhere among the answers, so we rescan until the chain reaches address records or ends. Hops are bounded by the answer count (in practice 0–4), so cost stays near-linear. The earlier revision performed the same bounded scans but read like accidental repetition, so this restructure is about clarity, not algorithm. Avoiding an index/map also keeps with the heapless model. Happy to flatten it if you see a cheaper shape with the same properties.

dns/client.go
hnw: Fair — I'll withdraw that. The hidden default is gone from `StartResolve`; since real-world responses do carry CNAME chains, `x/xnet` now passes an explicit minimal bound of 8 to stay on the safe side of typical CDN chains.

dns/dns_test.go
hnw: Trimmed — kept only regression coverage: the www.yahoo.co.jp case (compressed CNAME RDATA), reordered answers, and cyclic CNAME termination. Removed the synthetic "ignores non-address records" and decode-limit cases.

dns/client.go
```diff
@@ -24,9 +24,8 @@ type ResolveConfig struct {
 	Questions       []Question
 	Additional      []Resource
 	EnableRecursion bool
-	// MaxResponseAnswers limits how many answer records are decoded from the
-	// DNS response. If zero it defaults to the number of Questions.
-	MaxResponseAnswers uint16
+	MaxIPs          uint16
+	MaxCNAMEs       uint16
```
soypat: MaxAnswerIPs
MaxAnswerCNAMEs

dns/client.go
```diff
@@ -37,13 +36,14 @@ func (sudp *Client) ConnectionID() *uint64 { return &sudp.connID }
 
 func (c *Client) StartResolve(localPort, txid uint16, cfg ResolveConfig) error {
 	nd := len(cfg.Questions)
-	if nd > math.MaxUint16 {
+	if nd > math.MaxUint16 || nd == 0 {
```
soypat: good catch

soypat:
I played around with your PR a bit and reimagined the WriteAddresses functions- hope that's OK. Feel free to include my suggestions here or we can switch over to other PR, as you'd prefer 
https://github.com/soypat/lneto/pull/193

hnw:
Superseded by #193, which builds on these commits and adds @soypat's simplification of `WriteAnswers` (case-folded alias matching) and a single `MaxResponseAnswers` with CNAME headroom. Closing in favor of that PR — thanks for pushing it over the line, @soypat!

PR #190, opened 2026-08-24
tcp: simplified Policy implementation based on @MDr164

soypat:
@MDr164 I wanted to take a step in the direction of the API you've arrived at in https://github.com/soypat/lneto/pull/187

You'll note the scope is reduced at the tcp package level. I've ditched Nanotime entirely since I'm thinking that's a big fish to fry and will only be fully resolved if we fully redesign lneto.StackNode, which is on the roadmap since companies tell me that precise Rx time is necessary for them to adopt lneto.

Let me know what you think.


MDr164:
I'm a bit swamped with work this week but I'll take a look asap before the weekend. I also think that the nanotime stuff is inevitable and it enabled my TCP rework to implement a bunch of cool things but it's one of the larger breaking changes.

soypat:
I do tend to agree time will make its way in, but I'd like to imagine as much of the stack as possible without time before choosing at what layer to include it. I have a suspicion we can get away with leaving most of lneto free of time as part of the ***explicit*** API.

tcp/policy.go
```diff
@@ -0,0 +1,30 @@
...
+	// Reset returns the Policy to its pre-connection state. Called on every
+	// (re)open and Abort. Must preserve configuration such as a clock.
+	Reset()
+
+	// PreTx is called before writing to a frame.
+	// The outgoing frame options can be set by the Policy and will be respected if Frame offset >5.
+	// rtxFrom is ignored unless within [snd.UNA, snd.NXT] and returned retransmit==true.
+	// newTransmitLimit sets the maximum number of new bytes to send over the wire (congestion control).
+	// If not implementing congestion control then newTransmitLimit=[TransmitUnlimited].
+	PreTx(h *Handler, outgoingOpts Frame) (newTransmitLimit Size, rtxFrom Value, retransmit bool)
```
soypat: rtx->retransmit

tcp/policy.go
```diff
@@ -0,0 +1,30 @@
...
+	// PostTx called on leaving the transmit path with the fully written frame.
+	PostTx(h *Handler, outgoing Frame)
+
+	// PreRx is called by [Handler] on every incoming segment.
+	// PreRx can choose to drop segment if it returns keep=false.
+	PreRx(h *Handler, incoming Frame) (keep bool)
+	// PostRx is called by [Handler] after accepting an incoming segment.
+	// TODO: congestion control will also want the pre-Recv snd.UNA here.
```
soypat: Outdated todo

tcp/txqueue.go
```diff
@@ -331,6 +352,36 @@ func (sl *sentlist) Free() int {
 	return cap(sl.pkts) - len(sl.pkts)
 }
 
+// packetContaining returns the queued packet whose sequence range covers seq, or
+// nil when no packet does. It is the floor lookup a retransmission rewind needs:
+// seq lands inside a packet and the whole packet is resent.
```
soypat: Reword

tcp/txqueue.go
```diff
@@ -227,18 +227,39 @@ func (rtx *ringTx) RetransmitFromUNA() {
...
-			rtx.unsentend = rtx.sentend
-		}
-		rtx.sentoff = 0
+	rtx.RetransmitFrom(oldest.seq)
+}
+
+// RetransmitFrom rewinds the transmit queue so sent-but-unacked data at and
+// after seq becomes unsent again; the next MakePacket calls re-send it. seq is
+// snapped down to the start of the packet containing it — the retransmission
+// queue tracks whole packets, so sub-packet rewind is not representable. It is
```
soypat: Strong words...

tcp/policy.go
```diff
@@ -0,0 +1,30 @@
+package tcp
+
+// TransmitUnlimited size returned by [Policy.PreTx] to signal no new data transmit limit (no congestion control).
+const TransmitUnlimited = ^Size(0)
+
+// Policy observes segment traffic and steers transmit behaviour: RTO,
+// congestion control and the like (discussion #157). The tcp package holds no
+// clock, so a Policy needing time carries its own (issue #140).
+// Introspection stays off the interface; put it on the concrete type.
```
soypat: Clarify

tcp/policy.go
```diff
@@ -0,0 +1,30 @@
...
+type Policy interface {
+	// Reset returns the Policy to its pre-connection state. Called on every
+	// (re)open and Abort. Must preserve configuration such as a clock.
```
soypat: Should be call on every open

soypat (commented):
Looking at this from ams schipol passport control non-eu line so had a long time to take a look.

Note: loving memci output

PR #191, opened 2026-08-24
ci: laxify codecoverage

### LINKED ISSUE #185: Make code coverage CI pass requirements less stringent
Hitting 59% code coverage when repo coverage is at 60% should not fail. The requirements should be much lower.

I think it makes sense imposing stringent code coverage requirements when a PR is "large", but this is hard to measure. Maybe for now we can just turn it off since it is quite bothersome to have CI "fail" and be red because the percentage was slightly below what was needed.

soypat:
Solves #185


soypat:
So final effect is:
- No CI failure on patch diff. Can be 0%. Lax with PRs that make sound changes that may be 1-2 lines long but not covered by tests
- CI Failure if project coverage falls below 62%. This raises alarm on large PRs with low coverage which is the kind of PR we'd not enjoy

We may still encourage tests in small patches and also better prescribe writing of tests which often are written by LLMs and are just too many for small changes.

PR #192, opened 2026-08-26
xnet: debug packet before ingressing

soypat:
Debugging packet precedes ingress to catch packets which cause crashes within lneto. Although I've not explicitly observed a crash within lneto some behaviour with my LAN8720 server has made me more careful.


MDr164 (approved):

soypat:
Thanks for review @MDr164 !

PR #193, opened 2026-08-28
dns: @hnw alternate proposal

soypat:
Added
- simplify CNAME resolution and fold name comparison
- share ASCII case-folding helpers

### DIFF
--- a/dns/client.go
+++ b/dns/client.go
@@ -24,8 +24,11 @@ type ResolveConfig struct {
 	Questions       []Question
 	Additional      []Resource
 	EnableRecursion bool
-	MaxIPs          uint16
-	MaxCNAMEs       uint16
+	// MaxResponseAnswers limits how many answer records are decoded from the
+	// DNS response. If zero it defaults to the number of Questions. Answers
+	// are decoded in wire order regardless of type, so a response resolved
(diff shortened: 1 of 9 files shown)


soypat:
@hnw I was having a tough time figuring out what was bothering me in the function. Had a go at simplifying it and while I was at it I shared logic throughout lneto. Let me know what you think.

Also you'll note I went back on MaxAnswers. We can add CNAME/IP different decoding capacity in a separate PR

hnw:
Thanks for reworking this — I read through the full diff and I'm happy to adopt everything.

- Case folding for alias matching is a genuine fix over my version: DNS names are case-insensitive and the case of the CNAME target and the aliased owner names can differ within one response, so my byte-exact comparison could silently drop legitimate records. Adding `NamesEqualFold` to `Name` also gives this comparison a proper home for other call sites.
- `WriteAnswers` reads much better now — happy to keep it as is.
- Going back to a single `MaxResponseAnswers` with CNAME headroom (`len(addrbufnip) + 8`) is simpler than my `MaxIPs`/`MaxCNAMEs` split; separate capacities can wait for a dedicated PR if it ever matters.

Let's go with this PR — I'll close #189 in its favor. Thanks again!

soypat:
@hnw thank you for the review and for the improvement! Much appreciated!

PR #194, opened 2026-08-29
fix(xnet): don't fail DNS lookup when the address buffer is filled

hnw:
## Problem

Since #193, DNS lookups for hosts that answer with more A records than `addrbufnip` can hold fail with `resource exhausted`. Observed on real hardware (ESP32-C3 + espradio): resolving `www.google.com`, which currently returns eight A records, aborts with:

```
resolving www.google.com
ERROR: DNS: resource exhausted
```

#193 added CNAME headroom to `MaxResponseAnswers` in `StartLookupIPType` (`len(addrbufnip)+8` instead of `len(addrbufnip)`). This made `Message.WriteAnswers` reachable with more matching A records than fit in the caller-provided dst, which returns `lneto.ErrExhausted`. Before #193 the decode limit equaled the dst length, so this path was unreachable and untested. Hosts answering with five or more A records owned by the queried name (www.google.com: 8, support.apple.com: 6) hit it.

## Fix

`addrbufnip` is a fixed-size result buffer where filling it is a success, not a failure. When at least one address was written, treat `ErrExhausted` as success and drop the surplus addresses. No information is lost: callers that want to detect a surplus can already do so by comparing the returned length with their buffer size.

`TestDNS_TooManyAddrs` reproduces the failure mode with a response carrying eight A records for the queried name and verifies the lookup now succeeds, returning the buffered addresses.

Related: #189, #193 (the signalling role of `ErrExhausted` is unaffected for callers that pass a dst sized to their real needs; this only stops the fixed internal buffer in xnet from failing lookups it can legitimately satisfy).

### DIFF
--- a/x/xnet/stack-async.go
+++ b/x/xnet/stack-async.go
@@ -680,6 +680,11 @@ func (s *StackAsync) ResultLookupIP(host string) ([]netip.Addr, bool, error) {
 	if n == 0 && err == nil {
 		err = errDNSNoAns
 	}
+	if n > 0 && errors.Is(err, lneto.ErrExhausted) {
+		// addrbufnip holds a fixed number of addresses. Filling it is a
+		// success, so addresses decoded past its capacity are dropped.
+		err = nil
+	}
 	return s.addrbufnip[:n], true, err
 }
 
--- a/x/xnet/xnet_dns_test.go
+++ b/x/xnet/xnet_dns_test.go
@@ -239,6 +239,87 @@ func buildDNSMsgResponsePacket(t *testing.T, txid uint16, dstPort uint16, msg dn
 	return pkt, nil
 }
 
+// TestDNS_TooManyAddrs verifies that a response carrying more A records than
+// fit in the address buffer (addrbufnip) still resolves successfully with the
+// surplus addresses dropped, instead of failing with lneto.ErrExhausted.
+func TestDNS_TooManyAddrs(t *testing.T) {
+	const seed = 9876
+	const MTU = ethernet.MaxMTU
+
+	client := new(StackAsync)
+	dnsServerAddr := netip.AddrFrom4([4]byte{8, 8, 8, 8})
+	clientAddr := netip.AddrFrom4([4]byte{10, 0, 0, 100})
+	clientMAC := [6]byte{0xde, 0xad, 0xbe, 0xef, 0x00, 0x01}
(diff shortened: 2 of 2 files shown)


x/xnet/stack-async.go
```diff
@@ -680,6 +680,11 @@ func (s *StackAsync) ResultLookupIP(host string) ([]netip.Addr, bool, error) {
 	if n == 0 && err == nil {
 		err = errDNSNoAns
 	}
+	if n > 0 && errors.Is(err, lneto.ErrExhausted) {
+		// addrbufnip holds a fixed number of addresses. Filling it is a
+		// success, so addresses decoded past its capacity are dropped.
+		err = nil
+	}
```
soypat, line 687: I think this is pushing the problem somewhere else. We should allow users to own the memory so that they can pass in buffers they themselves sized which are written to. i.e: `ReadResultLookupIP(dst []netip.Addr, host string) (int,error)` and a `NumResultLookupIP() int` to accompany for users to size their buffer accordingly. the ErrExhausted buffer is important to be returned. Users should always be warned their buffer was insufficient. We don't sweep failures under the rug in lneto.

hnw:
Thanks for the review. Closing this PR — I understand the current
interface is intended as-is.

Just to confirm my understanding of the intended behavior after #193:

- Responses carrying more A records than `addrbufnip` can hold now
  return `ErrExhausted`, where they previously returned the first four
  decoded records. I understand this is intended, not a regression.
- Callers that only need a few addresses can proceed with the partial
  result, explicitly accepting `ErrExhausted`:

  ```go
  addrs, err := rstack.DoLookupIP(host, timeout, retries)
  if err != nil && !errors.Is(err, lneto.ErrExhausted) {
      fail(...)
  }
  // addrs holds the subset that fit
  ```

soypat:
1. Yes, ErrExhausted is pretty common throughout lneto. We ask users constrain their resources up front or explicitly decide for unbounded memory growth.
2. Yes. ErrExhausted is a warning error letting you know you need to increase the size of your buffer to fully capture the server response.

We should probably then fix the StackAsync implementation then:
- A well rounded fix would be to add a method that writes addresses to the user's buffer, so user decides to size buffer
- A weaker fix would be to add a configuration field that lets users size the addrbufnip dynamically on StackAsync configuration. But this approach kinda sucks because StackAsync is returning a slice it owns and can overwrite at any time. Not great...

Issue #201, opened 2026-09-13
DNS: DoLookupIP returns ErrInvalidAddr for A/AAAA responses containing CNAMEs, maybe fix or improve the errors


TheCoreMan:
Hey, new to this library (doing my first TinyGo project) so sorry if this issue isn't correct. Hopefully it is useful.

### Description

`DoLookupIP` fails with `invalid address` when an A/AAAA DNS lookup returns a CNAME record for the queried hostname.

I realize the README currently describes DNS support as:

> Client (A/AAAA query)

so transparent CNAME resolution may intentionally be outside the current scope.

However, the current failure mode looks misleading: it returns `invalid address`.

It would be useful if `DoLookupIP` either:

1. Transparently followed CNAMEs when resolving an IP so CNAMEs "just works", or
2. Explicitly reported an error that looks like "Hey, this is a CNAME, we don't support it".

### Reproducer

Using `lneto` indirectly through `tinygo-org/espradio`:

The `large...` domain is just a small relay server I've deployed to Koyeb for my side project. If you DIG it you'll see it's a CNAME the to `prod-glb...` domain:

```shell
❯ dig large-marylynne-thecoreman-172999c0.koyeb.app

; <<>> DiG 9.10.6 <<>> large-marylynne-thecoreman-172999c0.koyeb.app
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 36998
;; flags: qr rd ra; QUERY: 1, ANSWER: 3, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;large-marylynne-thecoreman-172999c0.koyeb.app. IN A

;; ANSWER SECTION:
large-marylynne-thecoreman-172999c0.koyeb.app. 300 IN CNAME prod-glb.koyeb.app.cdn.cloudflare.net.
prod-glb.koyeb.app.cdn.cloudflare.net. 300 IN A	172.66.172.174
prod-glb.koyeb.app.cdn.cloudflare.net. 300 IN A	104.20.31.27
```

```go
for _, host := range []string{
    "httpbin.org",
    "prod-glb.koyeb.app.cdn.cloudflare.net",
    "large-marylynne-thecoreman-172999c0.koyeb.app",
} {
    ip, err := wifi.GetHostByName(host)
    if err != nil {
[...]

PR #202, opened 2026-09-14
dns: CNAME hopping and canonical names

soypat:
logic driven mostly by LLM and reviewed by human.
prompt: issue 201 looks like we need to add a CNAME resolver layer in DoLookupIP.

Code looked mostly good on creation but will give it another review soon before marking ready for review. Seeks to solve #201 completely.


x/xnet/stack-blocking.go
```diff
@@ -158,12 +158,23 @@ func (s StackBlocking) DoLookupIP(host string, timeout time.Duration) (addrs []n
 // DoLookupIPType resolves host for the given record type (dns.TypeA or dns.TypeAAAA),
 // blocking until a response arrives or the timeout elapses.
 func (s StackBlocking) DoLookupIPType(host string, timeout time.Duration, qtype dns.Type) (addrs []netip.Addr, err error) {
+	const maxCNAMEHops = 3
```
MDr164: The naming isn't ideal on this one, or rather it's slightly wrong. I'd call it `maxCNAMEqueries` instead. Right now it actually just queries the first 3 entries and learns the fourth but never actually queries it. So if you'd expect to get 4 queries out of 3 hops you'd be deceived.

MDr164 (approved):
Generally good change, just had a minor naming item :)

x/xnet/stack-blocking.go
soypat: yeah, good catch. will change!

soypat:
This brings in breaking changes to DNS Lookup API to allow users to reuse DNS names and to prevent memory allocations during DNS CName hop querying. dns.Name is much more flexible in terms of fine grain memory control and can easily be created with `dns.MustNewName` so I feel it is worth the breakage.

Also worth mentioning I've refactored MWE to now include DNS+NTP in example so that we see the binary size increase which LLVM was previously DCEing.

soypat:
Will merge this soon to get the MWE memci improvements so future PRs have more accurate binary increase. Likely merge by wednesday unless something comes up that requires fixing/improving

examples/min-working-example/main-mwe.go
```diff
@@ -118,11 +129,11 @@ func run(ctx context.Context, stack *xnet.StackAsync) error {
 		time.Sleep(pollTime)
 		conn, err := listener.Accept()
 		if err != nil {
-			return makeMsgErr("listener.Accept failed", err)
+			err = makeMsgErr("listener.Accept failed", err)
```
MDr164, line 132: I think this error path is broken now. If `Accept()` returns an error then `conn` will normally be nil, but we still pass it to `handleConn` which immediately does `defer conn.Close()`. So this can turn an accept error into a nil pointer panic. I'd either return here like before or `continue` after storing/logging the error.

x/xnet/stack-async.go
```diff
@@ -614,14 +614,14 @@ var errNoDNSServer = errors.New("no DNS server- did DHCP complete? You can set a
...
-func (s *StackAsync) StartLookupIP(host string) error {
+func (s *StackAsync) StartLookupIP(host dns.Name) error {
 	return s.StartLookupIPType(host, dns.TypeA)
 }
 
 // StartLookupIPType begins resolving host for the given record type (e.g. dns.TypeA
 // or dns.TypeAAAA). The DNS query is always carried over IPv4 to the configured DNS
 // server; resolving over an IPv6 DNS transport is not yet supported.
-func (s *StackAsync) StartLookupIPType(host string, qtype dns.Type) error {
+func (s *StackAsync) StartLookupIPType(host dns.Name, qtype dns.Type) error {
```
MDr164, line 624: So imo the switch from `string` to `dns.Name` lost the synchronous input validation here. A zero `dns.Name` now makes it through `StartLookupIPType`/`StartResolve` and only fails later while encoding the packet, so the blocking lookup can end up timing out instead of returning an invalid-name error. Can we reject `host.Len() == 0` before starting the lookup? Or some other type of error checking probably. The new `dns.Name` type is nice but allows for a couple footguns compared to the simple string from before.

x/xnet/xnet_dns_test.go
```diff
@@ -339,3 +344,97 @@ func TestDNS_CNAMEResponse(t *testing.T) {
 		t.Errorf("expected address %s not found in result %v", wantAddr, addrs)
 	}
 }
+
+// TestDNS_CNAMEOnlyRequery verifies a response with only a CNAME is reported by
+// ResultLookupIP and followed by DoLookupIP with a query for the canonical name.
+func TestDNS_CNAMEOnlyRequery(t *testing.T) {
```
MDr164, line 350: Could we extend this test with a chain that actually reaches `maxCNAMEqueries` and ideally one that exceeds it? Right now it only tests the 2-query case, so the query-vs-hop distinction you just fixed in the naming isn't really covered and the final `errDNSOnlyCNAME` path seems easy to regress. So it kinda only tests the happy path if I see this correctly 😅

MDr164 (commented):
Found a couple things but otherwise seems good. So not too much left blocking a merge I'd say.

PR #203, opened 2026-09-16
TCP-NG core changes

MDr164:
Essentially part 1 of the tcp-ng changes from #187 with the largest change and core plumbing.


MDr164:
Stacked PRs only work if the branches already live in this repo, you can see what it would look like here: https://github.com/MDr164/lneto/pulls?q=is%3Apr+state%3Aopen @soypat

internal/ring.go
```diff
@@ -32,6 +32,14 @@ type Ring struct {
...
 	// If End==0 then the buffer is empty. If End==Off and End!=0 the buffer is full.
 	End int
+	// peeked is the number of bytes staged ahead of the write position by
+	// [Ring.PeekWrite] that have not yet been revealed by [Ring.Commit] (or
+	// overwritten by an in-order [Ring.Write]). It measures the distance from
+	// the current write position to the furthest staged byte. While it is
+	// positive a full read-drain must preserve the write position instead of
+	// resetting it, otherwise the staged bytes' physical offsets are invalidated
+	// (see [Ring.onReadEnd]). Used only by out-of-order TCP reassembly.
+	peeked int
```
soypat: Pretty sure we can create a new `RingPeek` type or something separate to Ring. I do enjoy segmenting logic and data structures as much as possible where it makes sense.

```go
type RingPeek struct {
    Ring
    peek int
}
```

soypat, line 42: This seems like a leak of an abstraction from TCP. The TCP reassembler should keep track of written data offset. The way I'd go about fixing this issue is two ways, both omit this field:
1. Add a check in Handler after reading all buffer data: if there is peeked data, then set Ring buffer offset to correct position
2. Remove the onRead check and never reset buffer, leaves responsibility to users. 

I like both of above. 1. leaks some Ring implementation details into TCP, but better than having every Ring user pay for peeks they do not use.
2. Ring becomes even more low level- however, users like TCP seem to keep track of buffer  availability/free so they may only need to do a check `buffered == n` to know they can call Reset and get the perf benefit.

tcp/rto/rto.go
```diff
@@ -142,22 +126,19 @@ func (r *Timer) NextDeadline() int64 {
 // PreRx keeps every segment: the estimator never drops traffic and records
 // nothing before the connection has decided whether the segment counts. It
 // implements [tcp.Policy].
-func (r *Timer) PreRx(h *tcp.Handler, incoming tcp.Frame) bool {
-	return true
+func (r *Timer) PreRx(rx tcp.RxMeta) tcp.RxDirective {
+	return tcp.RxDirective{Keep: true}
```
soypat, line 130: call me shirley but I am 99.999999999% true we'll omit RxDirective 🤷

tcp/policy.go
```diff
@@ -1,33 +1,313 @@
...
+// implementation that samples round-trip times from the timestamp option, or that
+// advertises selective acknowledgements, is not recovering from loss at all, yet
+// it needs exactly the same hooks at exactly the same points. Naming the seam
+// after one of its clients invited the assumption that the others did not belong.
+//
+// The core deliberately owns none of these. What stays in this package is the
+// RFC 9293 state machine, the sequence space and the buffers; what a policy adds
+// is timing, windows and options, none of which the state machine needs in order
+// to be correct. A connection with no policy at all still completes a handshake,
+// carries data and closes; it simply never retransmits.
```
soypat, line 19: tell ai to not touch comments unless it is fixing an error or updating stale documentation please

tcp/policy.go
```diff
@@ -1,33 +1,313 @@
...
+// RxDirective is returned by [Policy.PreRx].
+//
+// NOTE: its shape is the minimum viable contract — it mirrors the original
+// PreRx "keep" boolean from discussion #157 — and is the one element of the
+// interface not yet fully settled there. It is a struct (rather than a bare
+// bool) so fields can be added without breaking implementations.
+type RxDirective struct {
+	// Keep reports whether the received segment should be handed to the state
+	// machine. A false value drops the segment before it is processed.
+	Keep bool
```
soypat, line 312: I am definetely not a fan of these types. Tell AI to get rid of them and use current API. if AI resists take AI to a river or waterfront and comfort it, tell it everything will be fine. Tell it of all the comments it will be able to write once this PR is merged- and then purge its context window.

tcp/txqueue.go
```diff
@@ -215,48 +218,27 @@ func (rtx *ringTx) ring(off, end int) internal.Ring {
...
-		writeEnd = rtx.sentend
-	}
-	if rewindOff == rtx.sentoff {
-		rtx.sentoff = 0 // Whole queue rewound: sent region becomes empty.
-		rtx.sentend = 0
-	} else {
-		rtx.sentend = rewindOff
+// ok is false when start is not in the queued range [snd.UNA, snd.NXT), which
+// includes the case of an empty retransmission queue. The queue is not modified.
+func (rtx *ringTx) retransmitBoundary(start Value) (boundary Value, ok bool) {
```
soypat, line 235: We are replacing RetransmitFrom and 
EndSeq because we are exposing a lower level API, no?

tcp/control.go
```diff
@@ -78,6 +98,23 @@ type ControlBlock struct {
...
+	// Both shifts stay zero unless each side saw the option in the other's SYN,
+	// because scaling is only permitted when both offered it. The offer and what
+	// the peer offered are tracked separately since a passive open sees the peer's
+	// SYN before it sends its own option, and an active open the reverse.
+	sndWindShift   uint8
+	rcvWindShift   uint8
+	windShiftOffer uint8
+	windScaleSent  bool
+	peerWindShift  uint8
+	peerWindScale  bool
```
soypat, line 117: I believe we solved window scaling at the Handler level 👀

soypat (commented):
Was not able to take a look at all of it. I think I'd break it down muuch more. This is way too much for me to review and merge in a single PR. I'd like to also poke at the size of tests- it really seems like there may be more tests than is really necessary. When I have time I can take a shot at nibbling at issues brought up in smaller PRs

MDr164:
All fair comments. Let me try to get the policy/directive changes in line first. The rest is fairly uncontroversial imo and simply builds on top of this change. So let's try to figure that part out alone, I started in #204

MDr164:
Closing this again in favor of the smaller PRs that trickle in now. I'll keep the large draft open until congestion control is in though ;)

PR #204, opened 2026-09-17
Exclude policy options from advertised MSS

MDr164:
This is smallest step from the tcp-ng patch train I can pull out. Essentially paving the way for the policy/directive adaptions. Let me know if this looks ok @soypat


soypat (approved):
Oh snap, nice find

PR #205, opened 2026-09-19
dns: address panic in skip resource

soypat:
AI found a panic in skipResources. test panics with previous code.

AI:
The new test is TestDecodeMessageSkipTruncated in `dns/dns_test.go`. It covers a resource header cut short by 1–10 bytes, with both a zero-capacity and a nil destination. It also covers data longer than the message, and a complete skipped resource where the offset should reach the end of the message.

One behaviour change: mdns.Client passes nil for the authorities and additionals sections, so responses with malformed records there now return an error. They used to be silently accepted.


MDr164 (approved):
lgtm, tests pass and works locally for me

PR #206, opened 2026-09-20
Allow tracked retransmissions when packet queue is full

MDr164:
Another smaller fix pulled from the large tcp-ng pile of commits. I think it's better for everyone if I first extract small things like this before opening PRs that bundle several coherent changes or even API breakage.


MDr164:
I think the memci comment has some issues with my PRs, the output in the summary page work fine though. I'll take a look.

soypat (approved):
Thanks for improvement <3

PR #207, opened 2026-09-21
TCP: Retain duplicate-ACK recovery after local close

MDr164:
Another one from the pile. I think at some point we might need to rethink how we do testing here, it starts to not scale that well...


MDr164:
FYI, I'm about to run out of small things to push from the tcp pile, so there will eventually be a 1k LoC PR again soon-ish.

soypat:
Since the changes went to ControlBlock maybe we can find a way of expressing this test at the ControlBlock level? We do have table driven tests for ControlBlocks. And yes, we are overdue for organizing loose tests into table driven tests in many packages. One of these days...

soypat:
Looks like an improvement overall. That said I asked an LLM to review code and it came with interesting points:
- spurious dupack retransmit: After one peer closes connection and sends a FIN (which increases SEQ) the other peer can still remain half-open sending segments but not ack fin.
  - My opinion: I feel like this should not happen. It seems to me an lneto peer will ACK a FIN even if it half closes. I'm betting a dollar this is BS
- prexisting bug in Recv: Any ack acknowledges FIN and advances state.
    - My opinion: Opa, this sounds like it could be an issue. Unrelated to this PR.
- Remote triggerable panic: 
    - My opinion: Opa x10, and I was priding myself Lneto TCP was unpanicable. Have not verified.

These issues seem unrelated with this PR, so if you give me a thumbs up on this comment I'll go ahead and merge.

Full llm output for curious peeps:
<details>

**Verdict:** The fix is correct for the case it tests (data lost before our FIN, in FIN-WAIT-1). I checked it locally: the new test passes on the PR, fails on `main`, and the full `tcp` suite passes. But it introduces one new behavior in FIN-WAIT-1, and the Closing/LastAck half of the change can never run.

**1. New spurious "retransmit" when only the FIN is outstanding (introduced by this PR).**
Take the case where all data is acked and only the FIN is unacked, so `snd.UNA == snd.NXT-1`. Incoming ACKs with `ACK == UNA` still satisfy `IncomingIsDupACK`. That check has no `DATALEN == 0` requirement, so ordinary peer data segments during a half-close also count. After three of them, `HasPendingRetransmit()` becomes true. `Handler.Send` then emits a segment with `SEQ=UNA`, zero payload and no FIN, because `MakePacket` finds no data at UNA. I reproduced it:

```
client=FIN-WAIT-1, 3 dupacks -> pendingRtx=true
send after dupacks: SEG [ACK] ACK=1 SEQ=1 WND=1024   // no FIN, no data
```

It's mostly harmless, but it recovers nothing and bumps `nRetransmit`. On `main` this couldn't happen. A simple fix is to gate dupack counting and `HasPendingRetransmit` on actually having unacked *data*. Record the FIN's sequence number when it's sent, as a plain `Value` field with no heap use:

```go
// in Send, when hasFIN: tcb.finSeq = seg.SEQ + Value(seg.DATALEN)
func (tcb *ControlBlock) dataOutstanding() bool {
	return tcb._state.TxDataOpen() ||
		(tcb._state.txQueuedDataOpen() && tcb.snd.UNA.LessThan(tcb.finSeq))
}
```

I'd avoid the `UNA < NXT-1` shortcut because `RetransmitFrom` can rewind NXT below the FIN. Please also add a test for the FIN-only case.

**2. The Closing/LastAck coverage can't run (pre-existing bug in `Recv`).**
The state switch runs before the dupack check, and it treats *any* ACK as acknowledging our FIN:
- `StateLastAck`: any ACK calls `Abort()`. A dup ACK signalling lost data drops the connection and silently loses that data. I confirmed this: the state goes straight to CLOSED.
- `StateClosing`: any ACK moves to TIME-WAIT, again losing the data.

RFC 9293 §3.10.7.4 says to change state only if the ACK acknowledges our FIN, i.e. `seg.ACK == snd.NXT`, and otherwise ignore it. Until that's fixed, `txQueuedDataOpen()` only changes behavior in FIN-WAIT-1, so the PR description overstates what it covers.

**3. Remote-triggerable panic in TIME-WAIT (pre-existing, separate issue).**
`Recv` has no `StateTimeWait` case, so the `default:` branch panics. A retransmitted FIN is rejected earlier by the window check. An in-window pure ACK, however, passes validation and panics. I reproduced this on `main` with a normal active close followed by one pure ACK from the peer, and through the Closing path in #2. On a microcontroller, a spoofed in-window segment can crash the stack. It should get an explicit case: re-ACK and ignore.

**Nits**
- In `tcb.dupack < tcb.nRetransmit+retransmitMaxQueued+retransmitMaxQueued`, `retransmitMaxQueued` appears twice while the comment says `+retransmitAfterDupacks`. Either the code or the comment is wrong. This is pre-existing, but it's on a line the PR touches, so it's worth fixing here.
- The retransmit branch of `PendingSegment` doesn't cap `DATALEN` by `snd.MSS` or the window, unlike the normal path. This is pre-existing, but the PR widens where it applies.
- The test is good. It checks SEQ, flags and payload, that NXT is unchanged, and that the server reads the data. Consider asserting that the FIN is still pending or retransmittable after recovery.

</details>

MDr164:
The first is related to my PR here for sure and an oversight that I can fix here. The other two are unrelated to this PR as you mentioned already but still worth fixing. I'll provide fixes and will ponder a bit how to better detect panics, e.g. have a long-running fuzzing job that runs nightly for a couple hours or so.

MDr164:
OR we do the workaround of wrapping large chunks in `recover()` statements to catch panics and turn them into actionable errors. But I have no clue if that works as well on Tinygo as it does on regular Go.

tcp/handler.go
```diff
@@ -272,6 +272,11 @@ func (h *Handler) Recv(incomingPacket []byte) error {
 			// Update TX ring buffer to free up acked data.
 			h.bufTx.RecvACK(segIncoming.ACK)
 		}
+		if h.bufTx.BufferedSent() == 0 {
+			// FIN occupies sequence space but is not retransmittable payload.
+			h.scb.dupack = 0
+			h.scb.nRetransmit = 0
```
soypat, line 278: Be really great to represent this within ControlBlock- but I guess you tried/thought of that approach already and didn't work out?

MDr164: That's not **yet** easily doable. Currently ControlBlock knows about snd.UNA, snd.NXT, state, dupack and nRetransmit but not where the locally transmitted FIN sits in the sequence space. I already tried something like `dataOutstanding := snd.UNA.LessThan(snd.NXT - 1)` because the last sequence number should be the FIN but that breaks once the retransmission stuff is involved as `RetransmitFrom()` rewinds `tcb.snd.NXT = newNxt` while the already-transmitted FIN remains logically above the retransmitted data. Too much mental gymnastics in the morning.

I could extend TCB though to allow for this to be possible. I just couldn't get it working with what it currently controls.

soypat (approved):

tcp/handler.go
soypat: It's fine- I favor not adding more state when not explicitly needed- if the state needed is available in the handler and it makes relative sense for it to be there to implement this feature, then I'm ok with it. Was just a passing comment to see if there was an easy win on the table

PR #208, opened 2026-09-21
[CI TEST] Attempt to get memci comments from fork PRs

MDr164:
I think it's more GitHub weirdness and the memci comment only works if the PR originates from a branch from this repo. This jq mumbo-jumbo seemingly fixes it but I'm not proud of this change.


MDr164:
It think it won't pick up the yaml change from this PR before it's merged again...

soypat:
Hmmmm- this should probably live in memci instead of here, right?

LLM on memci integration
<details>

I checked both repos:

- `soypat/memci/comment/action.yml` has had the `commits/${HEAD_SHA}/pulls` fallback since its first commit (Sep 2).
- lneto already uses `soypat/memci/comment@main`.
- #208 was opened Sep 21 because fork PRs *still* got no comment.

So "go back to the tiny workflow with the commit→PRs fallback" is the setup that's broken. The likely cause is that `GET /repos/{base}/commits/{sha}/pulls` doesn't reliably return PRs whose head commit lives only in a fork. The `pulls?head=owner:branch` lookup in #208 exists for exactly that reason. It isn't redundant machinery.

Where #208 does overdo it:

- It inlines the whole action into lneto instead of fixing memci.
- The `--paginate | jq -s`, the `head.repo.full_name` filter, and the "exactly one" check aren't needed. `owner:branch` plus a SHA match is already unique.

The minimal fix is to change the lookup in memci and leave lneto's workflow as is:

```yaml
      env:
        # ...existing...
        HEAD_OWNER: ${{ github.event.workflow_run.head_repository.owner.login }}
        HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
      run: |
        # pull_requests[] is empty for fork PRs, and commits/{sha}/pulls on the
        # base repo misses fork-only commits. Look up by head ref instead.
        pr="${PR:-$(gh api -X GET "repos/${REPO}/pulls" \
          -f state=open -f head="${HEAD_OWNER}:${HEAD_BRANCH}" \
          --jq "map(select(.head.sha == \"${HEAD_SHA}\")) | .[0].number // empty")}"
```

This works for both same-repo and fork PRs, so it can replace the commits fallback rather than sit next to it. If the fork has pushed a newer commit since the run, the SHA won't match and the run skips commenting, which is what you want because the newer run will post.

One idea worth keeping from #208 is the `user.login == "github-actions[bot]"` filter when finding the comment to update. Without it, a comment from anyone that starts with the marker could get overwritten.

When you test this, remember that `workflow_run` always runs the workflow file from the default branch, as MDr164 noted in the PR. Neither #208 nor a memci change can prove itself from a fork PR until the change is on main. Push the memci fix, then open a throwaway PR from a fork.

</details>

PR #209, opened 2026-09-22
Cname fix: apply @MDr164 suggestions

soypat:
Created separate PR on #202 since didn't feel comfy making 202 even bigger than it already was and since I really like to look at these changes separately since I did not follow LLM completely in explanation

prompt: look at MDr164 suggestions in 202 and take a validation approach at lowest level using lneto.Validator on dns.Message to catch malformed hostnames synchronously


MDr164:
I know it's a draft but the code looked fairly coherent so I did a quick review: The new maxQueries test looks good now and the error path in main-mwe is also working again. The largest change is the validator here and I still see one small-ish issue: Datatype casting between uint16 and int.
In dns.go around line 704:
```go
func (r *Resource) Len() uint16 {
    return r.header.Name.Len() + 10 + uint16(len(r.data))
}
``` 
and then around line 389:
```go
l += int(m.Answers[i].Len())
``` 

So in `Resource.Len()` you can have an overflow; imagine RDLENGTH near the 65535 mark:
`name + 10-byte RR header + 65535-byte rdata > 65535`

I think we're full circle to my comment on the other PR that we might need to re-structure and clean up tests at some point to catch edge cases like those better. Either via fuzzing or something like https://github.com/hegeldev/hegel-go which I recently discovered and trying out for another project of mine. Or simply more table tests :)

PR #212, opened 2026-09-22
Handle incoming segments in TIME-WAIT

MDr164:
This is the fix for the remote panic issue found in https://github.com/soypat/lneto/pull/207#issuecomment-5772841279 and turns out this was even specified in the RFC so I added the comment where it originates from. Also added a regression test for this. The test is clanker-speak again though.


tcp/handler_test.go
```diff
@@ -21,6 +21,89 @@ func TestHandler(t *testing.T) {
 	sendDataFull(t, client, server, []byte("hello"), rawbuf[:])
 }
 
+func TestHandler_TimeWaitRecvACK(t *testing.T) {
```
soypat: I feel like this test should live in the tcp_test.go file along with other ControlBlock tests- and if we don't have any yet we can create a table driven test like we have for RFC exchanges but with expected error outputs

